PR #4, opened 2026-05-08
feat(ctl): ferrosa-ctl auth set-password subcommand

bkearns:
## Summary

Adds `ferrosa-ctl auth set-password --user <u> --realm <cql|graph> [--config <path>] [--force]` so the post-install prompt can seed the CQL admin user and the graph HTTP/Bolt basic-auth credential.

## CLI shape

```
ferrosa-ctl auth set-password --user admin --realm cql
ferrosa-ctl auth set-password --user admin --realm graph --force
ferrosa-ctl auth set-password --user admin --realm cql --config /etc/ferrosa/auth.yaml
```

- Reads the password twice from stdin without echo (`rpassword`) and errors on mismatch **before any disk write**.
- Existing `(realm, user)` entry prompts `[y/N]` before overwriting; `--force` skips the prompt.
- Output lands at `~/.ferrosa/config/auth.yaml` (overridable with `--config`), written atomically and chmod-0600 on Unix.
- Confirmation line on stderr, `process::exit(1)` on any failure.

## YAML layout

```yaml
cql:
  admin:
    hash: \"\$argon2id\$v=19\$m=19456,t=2,p=1\$...\"
    created_at: \"2026-05-08T20:30:00Z\"
graph:
  admin:
    hash: \"...\"
    created_at: \"...\"
```

Realms are merged: writing `cql/admin` never touches `graph/...` or other users in the same realm.

## Hashing scheme

**Argon2id**, OWASP 2024 parameters: m = 19 456 KiB, t = 2, p = 1, internal RNG salt (via `password-hash` 0.6).

Why Argon2id rather than matching the existing default? The CQL `PasswordAuthenticator` in `ferrosa-schema::auth::password` defaults to bcrypt cost 12, but its verify path (`PasswordHasher::verify_password_any`) **already auto-detects the \`\$argon2id\$\` prefix** and verifies through `argon2::Argon2::default()`. Argon2id is the OWASP-current recommendation and matches the second variant the schema crate already supports natively, so the new file format produces hashes that the existing verifier accepts without any server change to the hash format.

## Wire compatibility (CRITICAL for the installer)

The hash **format** is wire-compatible: a PHC `\$argon2id\$...` string written by this command is verifiable by `ferrosa-schema::auth::password::PasswordHasher::verify_password_any` exactly as-is. There is a regression test asserting this (`hash_is_wire_compatible_with_schema_verify_password_any`).

**However**, the running server does not currently read `~/.ferrosa/config/auth.yaml` at all. Its authoritative credential store is `system_auth.roles.salted_hash` (CQL table seeded by `seed_default_roles` in `ferrosa-schema/src/auth/bootstrap.rs`). So *as of this PR alone*, running `auth set-password` produces a correctly-formatted hash file that no server reads.

## Follow-up TODOs (intentionally NOT in this PR)

1. **Wire `auth.yaml` into bootstrap.** Teach `seed_default_roles` (or a new pre-bootstrap step) to read `~/.ferrosa/config/auth.yaml` and, when present, replace the well-known default `ferrosa_admin` / `ferrosa_user` hashes with the operator-supplied ones before the server accepts its first connection. Suggested location: `ferrosa-schema/src/auth/bootstrap.rs`, behind a config option pointing at the YAML path.
2. **Graph HTTP/Bolt path.** `graph` realm: confirm `ferrosa-graph` reads from the same source as the CQL realm or define an explicit graph-side reader.
3. **Password policy.** `auth set-password` currently rejects only empty passwords. Once bootstrap is wired, route the typed plaintext through `ferrosa_schema::auth::password::PasswordPolicy` so the installer enforces the same minimum strength as `CREATE ROLE`.
4. **Default-password warning.** Once bootstrap reads the YAML, `admin_password_is_default` should compare against the YAML-supplied hash rather than the hard-coded \`ferrosa_admin\` literal.

## Test coverage

26 unit tests in `ferrosa-ctl::auth::tests`:

- realm parsing (cql / graph / unknown)
- argon2id PHC format, verify-correct, reject-wrong, random-salt invariant
- **wire-compat** assertion against the schema crate's verifier path
- YAML round-trip and empty-realm omission (no spurious top-level keys)
- merge semantics: other realms preserved, other users in same realm preserved, same-user overwrite
- happy path: writes the hash to disk, stderr confirmation contains realm/user
- mismatch path: errors with `AuthError::Mismatch`, no file created, pre-existing file untouched
- existing-entry prompt: `n` returns `AuthError::Aborted` with original hash intact
- `--force` skips the prompt and overwrites
- empty password rejected
- `save` creates parent directories and sets mode 0600 on Unix
- 2 clap-parse tests in `main.rs` for the new subcommand (minimal + all flags)

Plus the existing 11 integration tests still pass.

## Test plan

- [x] `cargo test -p ferrosa-ctl` — 130 passed, 0 failed, 0 ignored
- [x] `cargo clippy --all-targets -- -D warnings` — clean across the workspace
- [x] `cargo fmt --all --check` — clean
- [x] `ferrosa-ctl auth set-password --help` renders correctly
- [x] `ferrosa-ctl --help` lists the new subcommand
- [ ] Manual smoke (post-merge): run against a real terminal to confirm `rpassword` hides the input
- [ ] Installer integration: invoke from the post-install script and confirm the resulting hash verifies via `verify_password_any` in a quick repl

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1810,6 +1810,8 @@ name = "ferrosa-ctl"
 version = "0.8.8+8"
 dependencies = [
  "arc-swap",
+ "argon2 0.6.0-rc.8",
+ "chrono",
  "clap",
  "crossterm",
  "ferrosa-cluster",
@@ -1817,11 +1819,16 @@ dependencies = [
  "ferrosa-schema",
  "ferrosa-storage",
  "ferrosa-udf",
+ "password-hash 0.6.1",
  "ratatui",
  "reqwest",
+ "rpassword",
+ "serde",
  "serde_json",
+ "serde_yaml",
  "tabled",
  "tempfile",
+ "thiserror 2.0.18",
  "tokio",
  "uuid",
 ]
@@ -4752,6 +4759,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "rpassword"
+version = "7.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "5ac5b223d9738ef56e0b98305410be40fa0941bf6036c56f1506751e43552d64"
+dependencies = [
+ "libc",
+ "rtoolbox",
+ "windows-sys 0.61.2",
+]
+
+[[package]]
 name = "rsa"
 version = "0.10.0-rc.16"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4768,6 +4786,16 @@ dependencies = [
  "signature",
  "spki",
  "zeroize",
+]
+
+[[package]]
+name = "rtoolbox"
+version = "0.0.5"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "50a0e551c1e27e1731aba276dbeaeac73f53c7cd34d1bda485d02bd1e0f36844"
+dependencies = [
+ "libc",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -5352,6 +5380,19 @@ dependencies = [
  "itoa",
  "ryu",
  "serde",
+]
+
+[[package]]
+name = "serde_yaml"
+version = "0.9.34+deprecated"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6a8b1a1a2ebf674015cc02edccce75287f1a0130d394307b36743c2f5d504b47"
+dependencies = [
+ "indexmap 2.14.0",
+ "itoa",
+ "ryu",
+ "serde",
+ "unsafe-libyaml",
 ]
 
 [[package]]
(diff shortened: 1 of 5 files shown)


bkearns:
Closing in favor of a different design — the file-based path doesn't fit Ferrosa's actual auth model.

**The miss:** Ferrosa stores password hashes in the `system_auth.roles.salted_hash` keyspace (Cassandra-compat), populated by `seed_default_roles()` at startup. The running server does not read `auth.yaml`, so the hashes this PR writes don't actually let anyone log in.

**The correct shape:** `ferrosa-ctl` already speaks CQL (it has `CqlClient` for the TUI). The right `auth set-password` is a thin wrapper around CQL `ALTER ROLE <name> WITH PASSWORD = '<plaintext>'` — the server then hashes via the existing `PasswordHasher` and writes to `system_auth`. ~50 lines of CQL client code, no bespoke hashing or file format needed.

The 26 unit tests, prompt-twice + `--force` semantics, and Argon2id parameter choices in this PR were genuinely good work — sorry to throw it out. The same agent has the codebase context loaded and is being re-briefed with the corrected approach.

Branch will be deleted; please use `feat/ctl-auth-set-password-cql` for the replacement PR.

PR #48, opened 2026-05-18
feat(storage): read_token_range — anti-entropy's missing primitive

bkearns:
## Executive Summary

Adds `StorageEngine::read_token_range(table_id, start_token, end_token, limit)` — returns all partitions whose tokens fall in `[start, end)`, up to `limit`.

Anti-entropy repair (#47) needs to answer "give me everything in this Merkle leaf's token sub-range," but the existing `read_range` is **key-bounded**: partition keys hash to tokens via Murmur3, so a contiguous token range is a discontiguous key range. Without this primitive, repair sessions read the first N partitions by key, filter to a token sub-range, and almost always get zero hits → algorithm runs correctly on the wrong inputs → no convergence on the wire.

## Implementation

MVP that materialises the full table up to `RANGE_READ_MATERIALIZATION_CAP` (10 000 partitions), then filters by token. Correct for any table at or under the cap — which covers the fmem-scale validation target (entity_store has ~9.8k partitions).

```rust
pub fn read_token_range(
    &self,
    table_id: &TableId,
    start_token: i64,
    end_token: i64,
    limit: usize,
) -> ferrosa_common::Result<Vec<Partition>>
```

Returns empty for `start >= end`, `limit == 0`, or unknown table. Delegates to a new `TableStore::read_token_range` that internally reuses `read_range_limited_rows` with the cap as the underlying limit, then applies the token filter.

A follow-up commit will add a streaming variant that walks the merged-stream iterator and short-circuits at `>= end_token` — that's what real production scale (10M+ partitions per table) needs. This MVP unblocks end-to-end repair without that larger refactor.

## Tests (4 new)

| Test | Asserts |
|---|---|
| `read_token_range_filters_by_token_bounds` | 20 partitions written; assert exactly the partitions whose tokens fall in the chosen range are returned, none outside |
| `read_token_range_empty_range_returns_empty` | `start == end` → no partitions |
| `read_token_range_full_range_returns_everything` | `[i64::MIN, i64::MAX)` returns all written partitions |
| `read_token_range_honors_limit` | limit caps the result count |

## Verification

- 669 ferrosa-storage tests pass (was 665; +4 new)
- `cargo clippy --all-targets -- -D warnings` clean
- `cargo fmt --check` clean

## Relationship to PR #47

Wire-up will happen in PR #47 (anti-entropy repair) — the repair-side code will switch from `read_range(None, None, N)` + filter to `read_token_range(start, end, N)`. That unblocks the fmem live-cluster validation.

## Test plan

- [ ] CI green
- [ ] Once #47 is rebased on top, run live `ferrosa-ctl repair --keyspace agent_memory --table entity_store` against the fmem cluster and verify per-node size convergence from 1346 / 383 / 207 MB toward roughly equal

### DIFF
--- a/ferrosa-storage/src/engine.rs
+++ b/ferrosa-storage/src/engine.rs
@@ -2108,7 +2108,7 @@ impl StorageEngine {
     ///
     /// Anti-entropy repair's per-Merkle-leaf "give me everything in this
     /// token sub-range" question can't be answered by the key-bounded
-    /// [`read_range`] — partition keys hash to tokens via Murmur3, so a
+    /// [`Self::read_range`] — partition keys hash to tokens via Murmur3, so a
     /// contiguous token range is a discontiguous key range. This primitive
     /// answers the token question directly by reading the merged stream
     /// once and short-circuiting at `end_token`.
--- a/ferrosa-storage/src/engine.rs
+++ b/ferrosa-storage/src/engine.rs
@@ -2103,6 +2103,35 @@ impl StorageEngine {
         self.read_range_limited_rows(table_id, start, end, limit, 0)
     }
 
+    /// Read all partitions whose tokens fall in `[start_token, end_token)`,
+    /// up to `limit`.
+    ///
+    /// Anti-entropy repair's per-Merkle-leaf "give me everything in this
+    /// token sub-range" question can't be answered by the key-bounded
+    /// [`read_range`] — partition keys hash to tokens via Murmur3, so a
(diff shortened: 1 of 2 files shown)


bkearns:
Closing as superseded. The `read_token_range` primitive this PR adds was landed via #47's squash merge (commit 741c9e4) with a better streaming implementation that scales memory with matches rather than table size. Diffing this branch against current main shows it would regress that streaming impl back to a materializing one — not the right direction. Follow-up work (#49, #50) continues from main.

PR #54, opened 2026-05-22
fix(compose): allow rustfs default creds for local-dev sample

zznate:
The root sample deployment binds rustfs to 0.0.0.0:9000 using the built-in rustfsadmin credentials. Recent rustfs builds refuse to start with default root credentials on a non-loopback listener, so `docker-compose up` aborts before the cluster can form.

Set RUSTFS_ALLOW_INSECURE_DEFAULT_CREDENTIALS=true on the rustfs service — the documented opt-in for local development, consistent with this sample already running with auth disabled.


bkearns:
Thanks @zznate I also moved some ports.  I've been using minio so that path isn't super well tested now.

PR #73, opened 2026-05-30
fix(cql,graph): C8 — multi-language driver compatibility (6 CQL drivers + Cypher/Bolt/SPARQL smoke)

bkearns:
## Summary

Completes **C8 — Full CQL Driver Compatibility** and re-enables the driver-smoke gate as **fail-loud**. All six language CQL drivers pass, the multi-protocol python suite (CQL + Cypher + Bolt + SPARQL + examples) is green, and an auth-enabled node now exercises auth enforcement.

Final `run-all.sh`: **7 passed, 0 failed**. Per driver: node 38/0, go pass, java 38/0, rust 44/0, csharp 42/0, python 123 passed/5 skipped, python-auth 4/4.

## Ferrosa CQL fixes
- **Collections**: bound `list`/`set`/`map` values decode via `types::decode_value` instead of being rejected as blobs.
- **NULL**: explicit `NULL` inserts write a column tombstone, not a live empty cell.
- **BATCH**: bound `?` values are substituted into batched statements.
- **TTL**: expired cells and fully-expired rows are filtered at read time.
- **v5 protocol framing**: gocql (legacy framing) and the DataStax Java driver (CRC modern framing) implement v5 framing incompatibly, so ferrosa caps negotiation at **v4** — one well-tested legacy transport for every driver. (v5 adds only perf/integrity framing; vectors are custom-type encoded and version-independent. Proper v5 modern framing tracked as future perf work.)
- **v5 query/EXECUTE flags**: read the 4-byte `[int]` flags width for v5.
- **PREPARE metadata typing**: type function calls (`COUNT(*)`→bigint) and `system.local`/`peers` columns so strict drivers unmarshal correctly.
- **LWT result shape**: an applied conditional returns only `[applied]`, not the full row.

## Graph / Bolt / Cypher fixes
- **Bolt auth**: send a proper Bolt `FAILURE` on invalid credentials instead of dropping the connection (drivers raise an auth error, not `SessionExpired`).
- **Cypher 2-hop WHERE**: predicates on hop-bound variables were pushed to the anchor scan and evaluated against unbound variables (dropping every row). They are now split into conjuncts and evaluated at the earliest point all their variables are bound.

## Test harness
- Add a second **auth-enabled** ferrosa node + `python-auth-tests` job so graph/Bolt auth enforcement is actually exercised; wire both into `run-all.sh`.
- Driver/test bug fixes: rust `SELECT *`→`(i32,)` arity, csharp map→`SortedDictionary`, java/go/python protocol + datacenter pins, python example double-brace/edge-label/import fixes.
- Auth-enforcement tests skip on the auth-disabled functional node; `cassandra_cql_examples` skips when the optional corpus is absent.
- **Re-enable driver-smoke fail-loud** + flip the workflow-guard test.

## Verification
- `ferrosa-cql` lib: 813 tests pass. `ferrosa-graph` lib: 301 tests pass. fmt + clippy clean.
- `tests/drivers/run-all.sh`: 7 passed, 0 failed end to end.


bkearns:
Follow-up: **#74** tracks closing the CQL **feature** gaps that the newly-vendored Cassandra example corpus surfaces (materialized views, role/auth syntax, ALTER KEYSPACE/TABLE, custom indexes, data masking, …), using those `.cql` examples as TDD red tests. This PR delivers **driver/language** compatibility; #74 is the language/feature-coverage backlog.

PR #91, opened 2026-06-07
fix: bound full-scan memory (intra-partition row streaming + coordinator paging) — OOM P0

bkearns:
## Problem (P0 — "OOM on full table scan")

A full-table `SELECT * FROM agent_memory.document_terms` (≈2.18M rows) OOM-killed cluster
nodes (cql-rt, anon-rss ~2.08 GB at the 2 GB cgroup cap). Root cause is **two independent
unbounded buffers** on the read path — the merged bounded-reader work (#83) bounded SSTable
*reader* memory but neither of these:

1. **Replica per-partition materialization.** The streaming range read bounded *partition
   count* (`STREAMING_CHUNK_PARTITIONS=64`), but a single `Partition` holds **all** its rows
   in one `Vec<Row>`. `document_terms` is an inverted index — one hot term is one partition with
   millions of postings; `StreamMerger::next_merged_partition` materialized it whole.
2. **Coordinator result accumulation.** The CQL FullScan path appended the entire result into
   `all_rows` before returning, ignoring server-side paging — so the coordinator buffered all
   2.18M rows. (Fixing #1 alone just moved the OOM from replica → coordinator, which the e2e
   caught.)

## Fix — bound BOTH buffers

**Commit 1 — intra-partition row streaming (`ferrosa-sstable`, `ferrosa-storage`, replica handler):**
- Bounded per-partition row reading (`next_partition_rows(limit)`) — never decodes a whole partition.
- A streaming **k-way row merge** (`PartitionFragment` / `next_fragment`) emitting ≤K-row fragments,
  preserving LWW / tombstone / static-row semantics incrementally — `O(num_sources + K)` resident
  regardless of partition width. Replica/coordinator chunk by total **rows**.
- `K = FERROSA_RANGE_READ_ROWS_PER_FRAGMENT` (default 4096).

**Commit 2 — coordinator server-side paging (`ferrosa-cql`):**
- Route the unbounded-scan shape through the existing `PagingState`: collect at most `page_size`
  rows, emit a `next_paging_state` resume cursor (raw partition+clustering bytes; inclusive
  re-seek + skip-≤-last) that resumes exactly-once **even mid-wide-partition**.
- Default `FERROSA_CQL_DEFAULT_PAGE_SIZE=5000` when the client sends none.
- LIMIT / ORDER BY / ANN / DISTINCT / COUNT / aggregate / predicate paths unchanged. Multi-replica
  token-merge-with-resume **fails loud** (no partial scans) — tracked follow-up.

## TDD

- **Storage**: `single_wide_partition_streams_in_bounded_fragments` (≤K resident); byte-identical
  equivalence vs whole-partition merge for LWW conflicts, tombstones, partition tombstones, static
  rows, mixed widths, and the cross-replica path, for k ∈ {1…1024}; existing streaming/digest
  equivalence oracles still pass.
- **CQL**: first page bounded by page_size; **paged traversal == whole scan (no gaps/dupes, incl.
  mid-wide-partition)**; default page applies with no client page_size; LIMIT/point reads unchanged.

## Verification

- CI green: `fmt --check`, `clippy --all-targets` (zero warnings), `cargo test -p ferrosa-sstable
  -p ferrosa-storage -p ferrosa-cluster -p ferrosa-cql`, `build --all-targets`.
- **E2E (the real proof):** re-ran the exact `SELECT *` over 2.18M rows on a 3-node cluster built
  from this branch. Before: all nodes → ~2.08 GB → OOMKilled. **After: RSS held flat (~200 MB
  per node; observed peak node1=304MB node2=499MB node3=277MB ), OOMKilled=false on all nodes.**
  _(Scan completion + final peak being confirmed; will update if anything changes.)_

### DIFF
--- a/ferrosa-cluster/src/coordinator/range_read_stream.rs
+++ b/ferrosa-cluster/src/coordinator/range_read_stream.rs
@@ -896,6 +896,81 @@ impl ClusterCoordinator {
         .await
     }
 
+    /// Resume-capable streaming range scan with an inclusive lower-bound key.
+    ///
+    /// Backs `WritePath::range_read_stream_all_from` (the coordinator-side
+    /// paging cursor). Only the local-only fan-out shape (CL=ONE with the
+    /// keyspace RF spanning the ring, i.e. no remote replica must be consulted)
+    /// is supported; any shape requiring a cross-replica merge is refused
+    /// rather than returning a partial scan, matching the unbounded scan guard.
+    pub async fn coordinate_range_read_stream_from(
+        &self,
+        table_id: &TableId,
+        start: Option<&ferrosa_common::key::DecoratedKey>,
+        cl: crate::consistency::ConsistencyLevel,
+        replication_factor: usize,
+    ) -> crate::error::Result<ClusterPartitionStream> {
+        let ring = self.ring.load();
+        let node_ids = ring.node_ids();
+        let node_count = node_ids.len();
+        let all_remotes_len = node_ids
+            .iter()
+            .filter(|&&id| id != self.local_node_id)
+            .count();
+        drop(ring);
+
+        let cl_remote_count =
+            remote_count_for_cl(cl, replication_factor, node_count, all_remotes_len);
+
+        if cl_remote_count > 0 {
+            return Err(ClusterError::Internal(
+                "paged cluster range scan with a resume key requires token-aware k-way stream merge across replicas; refusing to return a partial scan".into(),
+            ));
+        }
+
+        Ok(Box::pin(
+            self.storage
+                .range_iter_fragmented(table_id, start, None)
+                .map(|item| item.map_err(ClusterError::Storage)),
+        ))
+    }
+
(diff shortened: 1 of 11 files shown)


bkearns:
## E2E verification — PASSED ✅

Re-ran the exact `SELECT * FROM agent_memory.document_terms` (≈2.18M rows) on a 3-node cluster built from this branch (both fixes), monitoring all nodes throughout a multi-minute full scan:

| Node | Peak RSS (whole scan) | OOMKilled | Restarts |
|------|----------------------|-----------|----------|
| node1 | 304 MB | false | 0 |
| node2 | 499 MB | false | 0 |
| node3 | 277 MB | false | 0 |

**Before these fixes** the identical scan drove every node to ~2.08 GB (the 2 GB cgroup cap) and OOM-killed them within seconds. **After:** peak 499 MB, no OOM, no restarts. Memory is bounded by the per-fragment row cap + server-side page size, independent of partition width and total table size.

CI: all 11 checks green (Docs fixed in f1de3d12).

PR #129, opened 2026-06-14
fix(storage): close read-vs-compaction Ok(None) data-loss window (flaky concurrent_read_during_compaction)

bkearns:
## Root cause
A read snapshots the `ArcSwap` store view, then has its SSTable inputs compacted (view swapped + old files deleted) before it opens them → spurious `Ok(None)` = client-visible "row not found" during compaction = silent data loss. PR #104 covered only the partition path's open-error case; an investigation (workflow wrbm62kqp) found **three** residual windows. **Flaky code, not a flaky test** — the "reads must always be readable" assertion is correct and untouched.

## What this PR closes (2 of 3 windows + the demonstrated flake)
- **Window 3 — sibling read paths:** extracted a shared `with_retried_view()` helper and routed `read_limited_rows` **and the production hot path `read_clustering_row`** (CQL point reads via write_path.rs) through it. Previously only `read_limited_rows` retried.
- **Window 2 — Arc::ptr_eq false-negative:** dropped the pointer-equality gate; a stale-view open failure now retries by reopening against the current view (so S3 local-cache eviction re-resolves/re-fetches), bounded by `MAX_VIEW_RETRIES` and **fail-loud** if a fresh reopen still fails.

## Proof (deterministic, not luck)
- New deterministic test `clustering_read_during_compaction_retries_against_new_view` (ReadViewBarrier) — **fails before the fix, passes after**; existing `read_during_compaction_retries_against_new_view` still green.
- `concurrent_read_during_compaction` stress-looped **100× serial + 240× under core saturation, zero failures**; re-verified green on current main after rebase.

## Known remaining — tracked, NOT closed here (forge t_b367c166)
- **Window 1:** a missing `Filter.db` (optional component, `unwrap_or_default()`) deleted mid-open → degraded/empty bloom → key pruned → `Ok(None)` with no open-error, so the retry never fires. Plus a latent `num_bits==0` bloom panic. Needs `flush.rs` (fail-loud on absent Filter.db) + `ferrosa-sstable/reader.rs` (no `% 0`) — narrower window, fast-follow.

Gate: fmt/clippy/`cargo test -p ferrosa-storage`/workspace-doc clean. Same data-loss severity class as #125.

### DIFF
--- a/ferrosa-storage/src/engine.rs
+++ b/ferrosa-storage/src/engine.rs
@@ -12439,6 +12439,90 @@ mod tests {
         );
     }
 
+    /// Window (3): the same read-vs-compaction data-loss race, but exercised
+    /// through `read_clustering_row` — the production full-primary-key point-read
+    /// hot path (write_path.rs routes CQL `=` point reads here). Before the fix
+    /// this path had NO view-retry loop at all: a stale-view open failure just
+    /// `continue`d and the read returned a spurious `Ok(None)` = silent data loss.
+    /// It must now retry against the freshly-swapped view like the partition read.
+    #[tokio::test]
+    async fn clustering_read_during_compaction_retries_against_new_view() {
+        use crate::store::read_race_test_hook::{ReadViewBarrier, ARMED};
+        use std::sync::Arc;
+
+        let dir = tempfile::tempdir().unwrap();
+        let mut config = StorageEngineConfig::test_config(dir.path());
+        config.compaction.min_threshold = 2;
+        let engine = Arc::new(StorageEngine::new(config, None).unwrap());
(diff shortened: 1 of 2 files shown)


bkearns:
Follow-up: **window 1** (the third read-vs-compaction window listed here as "known remaining — tracked, NOT closed here, forge t_b367c166") is now closed in **#130** (branch `fix/read-during-compaction-race`, onto `main`). Since #129 was already squash-merged and its branch deleted, window 1 landed as a fast-follow PR rather than an update to this one. With #130, all three read-vs-compaction `Ok(None)`/panic windows are closed. forge task t_b367c166 is marked done.

PR #141, opened 2026-06-16
feat(install): install-memory.sh parity with install.sh (channels, idempotency, upgrade restart)

bkearns:
`docs/install-memory.sh` is the canonical installer served at ferrosadb.com/install-memory.sh (GitHub Pages). It lagged `install.sh` on updatability. This brings it to parity:

- `--channel stable|nightly` (matches install.sh; was stable-only)
- `.memory-version` stamp + "already up to date / use --force" short-circuit (was: always reinstalled)
- restart an already-registered launchd/systemd service on upgrade
- `--force`

Verified: `bash -n`, shellcheck clean, and live channel resolution (stable→v0.16.0, nightly→v0.21.0).

Note: a separate Homebrew tap (ferrosadb/homebrew-tap) now provides `brew install ferrosadb/tap/ferrosa[-memory]`.

### DIFF
--- a/.github/workflows/promote-release.yml
+++ b/.github/workflows/promote-release.yml
@@ -41,3 +41,22 @@ jobs:
           gh release edit "$TAG" --repo "$REPO" \
             --draft=false --prerelease=false --latest
           echo "Promoted ${TAG} to stable (latest, not prerelease)."
+
+      - name: Notify Homebrew tap to resync formulae
+        env:
+          # Fine-grained PAT (or classic repo-scoped token) with contents:write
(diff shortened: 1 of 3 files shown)


bkearns:
Updated: docs/install-memory.sh is now a **byte-identical mirror** of the canonical installer in ferrosadb/ferrosa-memory (PR #69), per the 'memory repo owns the installer' decision. It adds the guarded native `ferrosa-memory` setup CLI on top of the parity work. Merge order: land ferrosa-memory#69 first, then this, so the mirror matches the source.

PR #173, opened 2026-06-22
fix(onboarding): install harness hooks in the no-clone setup-memory.sh path

bkearns:
## The bug

Running `curl ferrosadb.com/setup-memory.sh | bash` and declining the source clone left onboarding **silently broken**: the binary installed, but the harness hooks (session-start / recall / turn-finalization — the whole point of memory) never did.

The only hook installer was the **repo-local `./setup.sh`** that `ONBOARDING.md` points to, which requires a checkout. So no-clone users hit a dangling reference and got no hooks. (Also: `ferrosadb.com/setup.sh` is a *different* script — the DB installer — a confusing name collision.)

## Fix

Add a hook-install stage to `setup-memory.sh` that runs on **every** path (default on; `--no-hooks` to skip):
- **with a checkout** → run `scripts/install-agent-hooks.py` in-tree;
- **without** → fetch the two self-contained, **stdlib-only** installer files (`install-agent-hooks.py` + `scripts/hooks/ferrosa-memory-turn-hook.py`) **pinned to the release `$VERSION`** into a stable dir (`$INSTALL_ROOT/share/ferrosa-memory`) and run them. The dir is stable because the generated wrappers bake in the hook path.

**Fail loud, never silent:** missing `python3` or a fetch failure prints actionable instructions + a WARNING that hooks weren't installed (server still runs). Dropped `--skip-auth-check` — it post-dates older release installers (would break the stable v0.15.0 path), and the installer already warns-and-continues when the server is unreachable.

New flags: `--hooks`/`--no-hooks`, `--harness <name>`, `--mcp-url <url>` (env `FERROSA_HARNESS`, `FERROSA_MEMORY_MCP_URL`).

## Validation
- `bash -n` + `shellcheck -S warning`: clean.
- **End-to-end against the live v0.15.0 release**: both hook files fetch (HTTP 200) and `install-agent-hooks.py` runs to completion with the script's exact flags.

Pairs with **ferrosadb/ferrosa-memory#<onboarding PR>** (ONBOARDING.md), same branch `fix/onboarding-no-clone-hooks`.

### DIFF
--- a/docs/setup-memory.sh
+++ b/docs/setup-memory.sh
@@ -39,6 +39,9 @@ VERSION=""
 WANT_CLONE=""    # ask|yes|no
 WANT_NOMIC=""    # ask|yes|no
 WANT_HERMES=""   # ask|yes|no
+WANT_HOOKS=""    # ask|yes|no — install harness hooks (default yes)
+HARNESS="${FERROSA_HARNESS:-auto}"   # auto|all|codex|claude|hermes|generic
+MCP_URL="${FERROSA_MEMORY_MCP_URL:-http://127.0.0.1:18765/mcp}"
 
 while [ $# -gt 0 ]; do
   case "$1" in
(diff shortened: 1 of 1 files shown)


bkearns:
Doc half: ferrosadb/ferrosa-memory#87 (ONBOARDING.md), same branch `fix/onboarding-no-clone-hooks`. Merge together.

PR #233, opened 2026-06-29
fix(cluster): stop stream-route close storm on sequence gap (paged-streaming ReadTimeout)

bkearns:
## What

The paged streaming range-read (which backs the viz edge stream and large `ALLOW FILTERING` scans) was stalling into `ReadTimeout` and serving partial snapshots. Root cause: a **stream-route close storm** in `stream_frame_router`, not clone/copy (the merge/forward path is move-based).

On an out-of-sequence chunk the router closed the route **and removed** the per-request seq state. A peer that keeps streaming after the close — or a `request_id` reused after a reconnect colliding with a stale in-flight stream — then had every straggler frame recreate fresh seq state (`next_chunk_seq = 0`) and re-close. One transient gap became an unrecoverable churn (`expected=0 observed=162,163,164…`) until the consumer idle budget elapsed → `ReadTimeout` → partial graph.

## Fix

Tombstone a `request_id` when its route closes on a sequence error; later frames drop silently. The read fails fast **once** and retries on a fresh monotonic `request_id` instead of churning. `request_id`s are monotonic per process, so the tombstone never needs resetting.

## Tests
- New `sequence_gap_closes_route_once_and_drops_straggler_frames`: a gap + 3 straggler frames closes the route exactly once (pre-fix: four).
- 1044 ferrosa-cluster lib tests pass; reconnect + Done-straggler-drain behavior unchanged; clippy/fmt clean.

forge t_0ae8b345.

### DIFF
--- a/ferrosa-cluster/src/coordinator/stream_frame_router.rs
+++ b/ferrosa-cluster/src/coordinator/stream_frame_router.rs
@@ -50,6 +50,18 @@ struct StreamSeqState {
 pub struct StreamFrameRouter {
     router: Arc<StreamRouter>,
     seq_state: Mutex<HashMap<StreamSeqKey, StreamSeqState>>,
+    // Requests whose route we closed on a sequence gap/mismatch. A still-sending
+    // peer keeps emitting chunks after the close; without this each one would
+    // recreate fresh seq state (`next_chunk_seq = 0`) via `entry().or_default()`
+    // and re-close — turning ONE transient gap (or a reused request_id colliding
+    // with a peer's stale stream after a reconnect) into an unrecoverable close
+    // storm (the `expected=0 observed=162,163,164…` churn that stalls the scan
+    // into a `ReadTimeout`). Tombstoned requests drop frames silently, so the
+    // read fails fast once and retries on a fresh (monotonic) request_id instead
+    // of churning until the consumer's idle budget elapses. request_ids are
+    // monotonic per process, so a tombstone never needs to be reset.
+    closed: Mutex<std::collections::HashSet<u32>>,
+    route_closures: std::sync::atomic::AtomicU64,
 }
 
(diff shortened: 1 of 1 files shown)


bkearns:
Deployed this branch to the local fmem-dev 3-node cluster and verified behavior under a real viz load.

**What this fix does (confirmed working):** the close *storm* is gone. Each colliding request_id now logs exactly **one** `sequence gap/reorder; closing stream route` and tombstones (stragglers drop silently) — node3 showed 213 distinct ids × 1 close each, then 1065 × 1 under heavier load. Pre-fix a single id spewed the climbing `observed_seq=162,163,164…` churn until ReadTimeout. So this PR correctly removes the CPU/log storm and is worth merging on that basis.

**What it does NOT fix (important):** the user-visible viz is still broken — typed-edge streams (all-scope + temporal, current & legacy-swapped) hit the 5s idle timeout and serve a partial snapshot; the WS never reaches `SnapshotStreamEnd`.

**The deploy revealed the real root cause, which is *not* what the commit message assumes.** It is **not** request_id reuse across restart. Every gap-close under load is *identical*: `expected_seq=0 observed_seq=5`, all from node1, on a **single stable connection**, with **fresh consecutive** request_ids (1072, 1073, 1074…). So for every fresh streaming read the router sees the first chunk at a constant seq=5 against fresh seq-state.

Leading hypothesis: node1 emits 5 chunks + a Done; `accept_done_seq` removes the (request_id, peer) seq-state on the matching Done; then continuation chunks (seq 5,6,…) for the same request_id arrive against fresh state → gap → close. A paged/continued-stream seq-lifecycle bug, tracked with full evidence in task t_dc729b1d (raised to P90 as the actual viz blocker).

Recommend: merge this for the storm mitigation, but keep #228/the viz issue open pending the seq-lifecycle fix.

PR #234, opened 2026-06-30
fix(cql): guard projected complex scans

bkearns:
## What

Adds the missing guard for projected complex full-scan SELECT shapes that could bypass PR #232's checked range-read arm.

Specifically, projected `DISTINCT` / `ORDER BY` / ANN full scans now probe at `DEFAULT_RANGE_READ_LIMIT + 1`; if more data exists, the router increments `range_read_truncated_total` and fails loud instead of materializing the whole projected table or computing over a clipped window.

Also tightens CI disk hygiene for the workspace-wide test jobs after GitHub Actions hit `rustc-LLVM ERROR: IO failure on output stream: No space left on device` on this PR. The test and cluster-integration jobs no longer restore the heavy `target` cache, free more preinstalled runner toolchains before compiling, and clean build artifacts with `if: always()` before post-job cache steps.

## Why

The PR #232 fix covered the non-projected complex scan arm via `range_read_limited_rows_checked`, but projection-eligible complex scans could still call `range_read_projected(..., None)` first. A query like `SELECT DISTINCT tenant FROM ...` over more than 10k partitions could therefore materialize all projected partitions.

## Tests

- `cargo fmt --check`
- `cargo test -p ferrosa-cql --lib` (973 passed, 1 existing ignored)
- `cargo test -p ferrosa-storage --lib range_read_truncated_increment_and_read -- --nocapture`
- `cargo run --quiet -p xtask -- p0-oom-audit --enforce --today 2026-06-30`
- `python3 scripts/check-unbounded-reads.py`
- `cargo clippy -p ferrosa-cql --all-targets --features test-util -- -D warnings`
- `RUSTDOCFLAGS="-D warnings" cargo doc -p ferrosa-cql --no-deps --features test-util`
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/ci.yml"); puts "ci.yml parses"'`
- `actionlint .github/workflows/ci.yml`

Note: plain `cargo clippy -p ferrosa-cql --all-targets` hits the existing `protocol_conformance.rs` `test-util` feature gate, so the passing clippy command enables that feature.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,9 +73,22 @@ jobs:
       - uses: Swatinem/rust-cache@c19371144df3bb44fab255c43d04cbc2ab54d1c4 # v2.9.1
         with:
           cache-on-failure: false
+          # This job builds the whole workspace with --all-features. Restoring a
+          # previous target dir can consume enough runner disk that rustc fails
+          # while writing fresh artifacts; rely on sccache for compiler reuse.
+          cache-targets: false
       - name: Free disk space
         run: |
-          sudo rm -rf /usr/share/dotnet /usr/local/lib/android /opt/ghc || true
+          df -h /
+          sudo rm -rf \
+            /usr/share/dotnet \
+            /usr/local/lib/android \
+            /opt/ghc \
+            /usr/local/.ghcup \
+            /usr/share/swift \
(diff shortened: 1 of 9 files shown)


bkearns:
We aren't going to do this, this is the wrong approach this puts caps in vs streaming.

PR #237, opened 2026-07-01
feat: stream range reads — remove the server-side result cap (no DEFAULT_RANGE_READ_LIMIT)

bkearns:
**Status:** final live-cluster gates running; auto-merge will be armed when they pass. Stacked under this: #238 (audit rules).

## Principle
A query's result is bounded **only** by its own `LIMIT`/paging (or ANN top-`K`). The 10k `DEFAULT_RANGE_READ_LIMIT` server-side result cap is removed for every streamable shape. Safety comes from bounding **memory** (streaming + bounded buffers + spill), never result count. Closes the direction behind the rejected #234; supersedes #232. Design: `specs/proposed/streaming-range-reads-no-cap.md`.

## What this PR contains (in order landed; each implement + adversarial-verify)

**Streaming shapes (steps 1/2/5)**
1. **Projected arm → coordinated streaming** (`34576b1e`) — also fixed a real cluster bug: the old projected path read only the local node (silently partial on RF>1).
2. **Uncap O(1)-streamable shapes** (`a803e421`) — scalar aggregates (previously refused) fold through O(1) streaming accumulators; user `LIMIT N > 10k` streams `.take(N)`; `DISTINCT <pk>` locked uncapped; 10k clamps removed from `write_path`/coordinator/raft.
3. **Unbounded `ORDER BY` → spill** (`f08094a7`) — external merge sort (`external_sort.rs`) + RAM-budget spill threshold (`spill_budget.rs`, cgroup-aware, 50% default, tunable). Complete, correctly ordered, memory bounded; fail-loud on spill I/O.

**Coordinated projected paging: attempted, live-falsified, reverted to fail-loud (`2de9097a`/`89215318`)**
The partition-limit + producer-cancel implementation (`36140808`/`3ec30240`) passed all in-process tests but hung + OOM'd the real RF=3 cluster (abandoned remote producers). Both commits are reverted: a cluster projected scan with a page bound now **fails loud** ("not implemented; refusing to return partial results") instead of hanging. The real fix needs a genuine multi-node harness — plan in `specs/proposed/multi-node-streaming-test-harness.md` (fly.io scaffold in `deploy/fly-stream-scan/`, not CI-wired), tracked `t_3fc6be3c`.

**Bounded coordinator consume path (`ffa9325c`)**
`consume_range_stream` accumulated every partition from every replica into a `Vec` (peak = O(result) — the OOM root shared by the fts incident). Now partitions move through a `PartitionSink` over a bounded mpsc (backpressure); the limited-rows coordinator drinks the token-deduped N-way merge and folds ≤ limit partitions. Alloc-measured proof: consumer peak flat 284 KB→291 KB as N grows 750→12,000 (ratio 1.02); reverting the fix: ratio 8.59. Faithful RED harness in `tests/replica_scan_serialization_memory_bound.rs`.

**fts_match OOM, both layers (`f587808d`, `05b0b205`) — t_ee98faa0**
A single `hybrid_search` FTS query OOM-killed live nodes at the intentional 2 GiB cap.
- *Coordinator layer:* the fts arm materialized ALL matched rows before applying LIMIT → early-exit inside the fetch loop (post-filter counts, geometric escalation with a proven completeness invariant), deterministic key ordering, no-LIMIT queries page with a cursor (complete results, bounded memory).
- *Replica layer:* `fulltext_search` deserialized whole FTI sidecars + scored ALL postings per node → query-derived top-k pushdown end-to-end (CQL arm → RPC `limit` field → engine → streamed sidecar scan `scan_term_top_k` + bounded heap). Engine-level peak: 45 MB scaling ×16 with match count → **9.4 KB flat**. Only the queried index is touched (tested, incl. dangling registrations).
- ⚠️ **Deploy note:** `FulltextSearchRequestPayload` gained a field (bincode) — mixed-version clusters can't interop on this RPC; restart all nodes together.

**Hygiene:** stream-named fns returning bounded Vecs renamed (`42a6022a`, `32f9e5cf` — audit naming contract); flaky wall-clock test made deterministic (`dddd704e`); pre-existing `#[ignore]` fts test un-ignored and passing — ferrosa-cql has **zero ignored tests**.

## Live-cluster validation (3-node RF=3, 2 GiB mem caps — the caps are intentional and unchanged)
- The exact killer query (`context_snippet = fts_match('memory') LIMIT 10` over real `entity_store`) — previously OOM-killed all 3 nodes — returns correct rows, zero restarts, memory flat. ✅
- End-to-end `hybrid_search` through the live MCP: `total_candidates: 25` (was 0 + dead node). ✅
- `COUNT`/`SUM`/`MIN`/`MAX` exact over fresh 15k-row table. ✅
- Remaining gates in flight: full streaming battery + viz `SnapshotStreamEnd` probe — auto-merge armed on green.

## Still guarded / tracked follow-ups
Coordinated projected paged scans fail loud pending the multi-node harness (`t_3fc6be3c`); no-LIMIT fts returns the complete doc-key set (streaming RPC deferred, `t_112bb4e5`); non-Term fts shapes deserialize one sidecar at a time (deferred); GROUP BY spill when GROUP BY parses (`t_2b30130b`); Vec-at-caller wrappers for SPARQL/graph/CQL fallbacks (`t_480c91fe`); prepared `LIMIT ?` bind not substituted — pre-existing (`t_e4b830c7`); DROP TABLE dangling index registrations found during this work (`t_ae06e925`, fix in flight); CQL v5 modern-framing CRC (unrelated, `t_c7031765`).


bkearns:
## ⚠️ Live-cluster validation — merge-blocker found (do not merge yet)

Validated on the 3-node fmem-dev cluster (RF=3, 15,000-row table + real typed_edges).

**Works on the cluster (the hard/risky parts are proven):**
- `COUNT(*)` = 15000 exact · `SUM/MIN/MAX` exact · `DISTINCT` = 15000 complete
- **`ORDER BY` (no LIMIT) = 15000 complete + correctly sorted via the step-5 spill path** ✓
- **No OOM** (nodes ~360 MB / 2 GB, `OOMKilled=false`), no data loss, no miscount, no silent partials — every unsupported shape fails *loud*, never returns wrong data.

**Blocker (`t_fed055cb`):** a **paginated projected `SELECT <cols>`** (the most common shape — pages by default) and `SELECT <cols> LIMIT N` **fail-loud** on the coordinated cluster path:
`projected cluster range scan with partition_limit is not implemented`.
This is the step-1 follow-up, now confirmed as a hot-path shape (not a corner case). The unit tests passed because they only exercise the single-node **Direct** path.

**Fix before merge:** implement the coordinated projected partition-limit (apply the bound across scatter-gather + paging), with multi-node cluster tests. Holding this PR until that lands and re-validates green on the live cluster.

bkearns:
## ⛔ Live re-validation found a NEW blocker — still do not merge

After the coordinated-projected-partition-limit fix (36140808) + CI fixes, re-validated on the live 3-node RF=3 fmem-dev cluster. CI is green, but the cluster caught a regression the single-node unit test could not:

**A multi-page projected scan HANGS on a multi-replica cluster.** On a *fresh* table (15k rows, no restart): `COUNT(*)`=15000 ✓, `SUM/MIN/MAX` exact ✓, `DISTINCT`=15000 ✓, `SELECT src_id LIMIT 100` (1 page) ✓ — but `SELECT id` (15k → 3 pages) **times out** (`defunct by heartbeat`). `SELECT id,v`, `SELECT id LIMIT 15000`, `ORDER BY v` all hang/return-0 on RF=3.

- **Data is intact** (SUM exact over all rows, DISTINCT complete) and **no OOM** — it's a coordinator **paging** bug, not data loss or memory.
- Root cause: the partition-count-limit paging works on the local-only fan-out (`expected_done==0`, all the unit test exercised) but the **multi-replica fan-out (`expected_done>1`) doesn't resume across pages** — page 2's `paging_state` continuation deadlocks. Worse, 36140808 turned the prior *fail-loud* into a *hang*.

Tracked as **`t_3fc6be3c`** (P92, blocking). Fix requires a real multi-node (RF>1) test harness — the single-node harness structurally cannot reproduce it. Interim option under consideration: revert 36140808 to fail-loud (fail-loud > hang) until the coordinated paging is correct.

The streaming wins (aggregates, DISTINCT, COUNT, ORDER-BY spill, no-OOM) all hold on the cluster; the projected paged-scan coordinated path is the remaining gap.

PR #244, opened 2026-07-03
fix(read-path): correct COUNT(*)/paged-scan under-count + compaction resilience for legacy SSTables

bkearns:
## What & why

Full-table `COUNT(*)` and paged `SELECT` scans of a large real table (`agent_memory.typed_edges`) returned **non-deterministic, under-counted** results (viz showed ~10–16k of the true 21,168). Root-caused to two independent read-path bugs plus a data-at-rest issue, then hardened so the whole class fails loud instead of silently under-delivering.

Live-validated on the fmem-dev 3-node cluster: `COUNT(*) = 21168`, and paged scans return the full `21168` at fetch sizes 500 / 5000 / 50000, deterministically.

## The bugs

- **Bug A — `COUNT(*)` over-count.** The range-merger fused token-disjoint runs on SSTables whose bounds are **not byte-comparable-decodable** (legacy Cassandra-format files), double-counting `(pk, clustering)` rows across sources. Fixed by decode-guarded run grouping (non-decodable bounds → singleton run).
- **Bug B — paged scan under-delivers ~half a wide partition.** A legacy SSTable stores a wide partition's rows in an order that differs from the raw-clustering comparison the streaming fragment merge + paged resume filter use, so the partition is emitted as **two concatenated ascending runs**. The resume filter (`clustering > resume_ck`) then silently drops the second run across pages. The real fix is at rest (compaction, below); the read path is now made **fail-loud** so it can never silently under-deliver again.
- **Bug C — corrupt on-disk timestamps.** Out-of-range `created_at` values broke `SELECT *` decode; writes now reject out-of-range timestamps and the read path fails loud on corrupt on-disk values.

## Resilience (the durable fixes)

- **Two fail-loud read guards (no buffering):**
  - `range_merger::emit_fragment` errors on out-of-clustering-order rows within a partition (full-read path).
  - `write_path::resume_filtered_stream` errors on non-monotonic delivery of the resume partition (paged path — where the silent drop actually happened). Both are inert on healthy data.
- **Legacy-format rewrite (compaction fixes it at rest).** `SSTableMetadata::legacy_format` (computed from reader bounds) + `strategy::legacy_rewrite_tasks`, wired into `maybe_compact`: any non-byte-comparable SSTable is scheduled for a rewrite **regardless of size tier**, so size bucketing can't leave a lone legacy file lingering. Compacting through `merge_partitions` re-sorts each partition and writes byte-comparable output — permanently correct, and every read path stays O(k).
- **Memory/CPU-tuned compaction parallelism.** Compaction now streams a partition-group at a time (per-task memory ≈ widest partition × inputs, not the whole SSTable), so the old fixed cap of 2 concurrent merges was stale and starved a deep backlog. Concurrency + worker count now auto-derive from cgroup/system memory and CPUs (`min(cpus, mem/2 ÷ 256 MB)`, clamped 1–8); env vars still override. On a 2 GB node: 4 concurrent (up from 2), peak compaction memory held under half the limit.

## Live validation (fmem-dev, this branch's image)

- All 3 nodes healthy on the new image; auto-tune logs `worker_count=8 max_concurrent=4` on each.
- `COUNT = 21168`; paged scan `21168` at fs=500/5000/50000, deterministic; resume guard inert (no false positives).
- The last legacy-format SSTable (`decode=false`) was compacted away → single consolidated SSTable with `decode=(true,true)`, `21168` rows, `inversions=0`.

## Testing

- `cargo fmt --check`, `cargo clippy --all-targets`, `cargo test --workspace` — all green locally.
- New unit tests: range-merger monotonic-merge + out-of-order fail-loud; resume-stream fail-loud + monotonic prefix-drop; `legacy_rewrite_tasks` (lone/multiple/none); `auto_tuned_max_concurrent` (cpu+memory bounds) + workers.

## Follow-ups (tracked)

- Startup / periodic legacy-format sweep so cold (read-only) tables also self-heal (t_1d235673).
- Targeted per-table force-compact so an operator intent doesn't queue behind every table (t_eacb1390).


bkearns:
Both review findings fixed in a8addd48.

**1 (blocking) — single-source rewrite didn't re-order rows.** Confirmed: `merge_partitions` short-circuits `sources.len() == 1` and returns rows unsorted (merge.rs), so a 1-input legacy rewrite kept its non-monotonic order while stamping `legacy_format=false`. (Worse: any partition present in only one input of a multi-input task was also left unsorted.)

Fix: extracted the sort+dedup into `merge::sort_and_dedup_rows` and added `merge::ensure_partition_rows_sorted`, which the compaction executor now calls on **every** merged partition — so all compaction output is monotonic regardless of source count. It's guarded by an O(n) `is_sorted_by` check, so it's free for the already-sorted common case (every multi-source merge, every modern file), and I did **not** touch `merge_partitions`' single-source fast path, so read paths pay nothing.

Regression test (your exact ask): `compaction_resorts_single_misordered_legacy_sstable` writes a single SSTable whose rows are stored `[5,1,2,3,4]` (asserted as a precondition), compacts that one input, and asserts the output reads back `[1,2,3,4,5]`.

**2 (high) — unbounded single legacy task + all-or-nothing overlap skip.** Fix: `legacy_rewrite_tasks` now chunks by the same `max_threshold` / `max_compaction_bytes` bounds, each chunk a **separate** task (an input-overlap skip drops only that chunk, not the whole repair set); a single oversized legacy file still gets its own task rather than being dropped. And `maybe_compact` now excludes legacy files already covered by a strategy task, so no overlapping repair task is submitted (the strategy task rewrites them sorted anyway via fix #1). Tests: `legacy_rewrites_chunk_by_count_and_size_bounds`, `oversized_lone_legacy_file_still_gets_its_own_task`.

Full storage lib green (968). fmt/clippy clean.

PR #265, opened 2026-07-16
fix(fulltext): stream fts_match keys end to end with bounded memory; delete the geometric escalation loop

bkearns:
## Problem

A broad no-`LIMIT` `fts_match` (reached directly or via the router's geometric LIMIT escalation) materialized its match set several times over: every replica scored all matches into a `score_map`, shipped one bincode response frame of up to 256 MiB, and the coordinator+router held the union in ~3 more copies. Concurrent scans OOM-killed all three 2 GiB-capped nodes in cascade (forge t_8fc24ce2); the blocking FTI walk also ran inline on the async runtime, starving raft heartbeats/CQL keepalives and turning Bulk-lane timeouts into a retry storm that multiplied the buffers.

## Fix (forge t_4ae47a9f) — streaming, non-materializing, backpressured; no result caps

- **ferrosa-index**: `scan_term_each` — non-materializing postings walk (O(1) working set, `ControlFlow` early-exit); `scan_term_top_k` reimplemented on it.
- **ferrosa-storage**: `fulltext_search_each` — streaming no-LIMIT walk over memtable + sidecars + missing-sidecar fallback; the walk holds no score map, and blocking scans now run via `spawn_blocking` at every call site (same class as PR #131).
- **ferrosa-net**: additive `FulltextSearchStream{Request,Chunk,Heartbeat,Done,Cancel}` wire family (0x3B–0x3F), ordered-stream classified like the ADR-020 range-read stream.
- **ferrosa-cluster**: producer walks the FTI on a blocking thread into a bounded channel and fires ≤4096-key chunks with heartbeats + cancel; `coordinate_fulltext_search_stream` fans out to every node, N-way merges over bounded channels, and dedups into a single `seen` set — the one intentional O(distinct matches) allocation left. Any replica failure fails the search loudly (no silent partial union). `FERROSA_BULK_STREAMING_FULLTEXT=0` selects the legacy path for mixed-version upgrades.
- **ferrosa-cql**: the `fts_match` arm consumes the stream for no-LIMIT; LIMIT-k keeps one bounded top-k consult with at most **one** complete-stream fallback — **the geometric ×4 escalation loop is deleted** (its terminal round degraded to `limit=None` materialization on every replica). Paging semantics unchanged.

## Verification

- `ferrosa-storage/tests/fulltext_streaming_each_memory_bound.rs`: allocator-tracked walk peak < 1 MiB, independent of a 16× match-count increase; parity/early-exit/fail-loud contracts.
- `ferrosa-cluster/tests/fulltext_stream_memory_bound.rs`: e2e completeness at N=8000 through the cluster write path, peak independent of a 64× doc-size increase, 256 B/key backstop, early-drop teardown.
- New structural router pin: hostile post-filter ⇒ exactly 2 hit-set fetches; no-LIMIT ⇒ 1 (a regression to the loop fails the pin).
- Full local gates: fmt check, clippy all-targets (0 findings), the whole workspace suite (only the pre-existing env-gated ferrosa-jepsen container tests fail without `FERROSA_TEST_CONTAINERS`), and the docs build with `-D warnings` — all clean.
- Multi-node fly.io validation staged (deploy/fly-stream-scan `fts_content_scan` probe is this exact query shape); tracked on t_4ae47a9f.

## Notes for reviewers

- Design delta from the original spec: no producer-side window/credit protocol — FTI postings can't resume by position without a sidecar format change; the producer's bounded internal channel already gives consumer pacing, and in-flight exposure is ≤160 KiB frames × route buffer. The e2e guard covers this; if a stalled-consumer profile ever shows growth, the credit-frame refinement is the follow-up.
- The streaming path is stricter than the legacy degrading union: a down replica now fails the query loudly instead of silently dropping its matches.

### DIFF
--- a/ferrosa-cluster/README.md
+++ b/ferrosa-cluster/README.md
@@ -100,6 +100,18 @@ strict-serializable multi-key / cross-shard transactions and LWT.
   leader-only serial, cross-DC Accord routing).
 - `coordinator/batch.rs` — 3-phase logged batchlog (write → fan out → delete only
   on full success) with replay task; `DEFAULT_BATCH_CONCURRENCY = 32`.
+- `coordinator/fulltext_stream.rs` — streaming fulltext search (t_4ae47a9f),
+  the `fts_match` twin of the ADR-020 range-read stream: the producer walks the
+  local FTI via `fulltext_search_each` on a blocking thread and fires bounded
+  `FulltextSearchStreamChunk` key batches (≤ 4096 keys) on `Lane::Bulk` with
+  heartbeats + Cancel; `coordinate_fulltext_search_stream` fans out to every
+  node, N-way merges over bounded channels, and dedups into one `seen` set —
+  the only O(distinct matches) allocation left in the path (scores and extra
+  copies are gone; the pre-fix union OOM-killed replicas, t_8fc24ce2). Any
+  replica failure fails the stream loudly — no silent partial union (stricter
+  than the legacy degrading path). `WritePath::fulltext_search_stream` is the
+  mode-dispatching entry; `FERROSA_BULK_STREAMING_FULLTEXT=0` falls back to
+  the legacy single-message union for mixed-version upgrades.
 - `coordinator/{range_read_stream,stream_*}.rs` — ADR-020 streaming range reads
   and projected streaming scans; the old Vec-returning
   `WritePath::range_read_projected` wrapper has been removed, so projected
--- a/ferrosa-cluster/src/controller/cluster.rs
+++ b/ferrosa-cluster/src/controller/cluster.rs
@@ -924,6 +924,7 @@ impl ModeController {
         // on the next line and is no longer directly accessible.
         let stream_router_for_handler = coordinator.stream_router();
         let peer_manager_for_handler = coordinator.peer_manager.clone();
+        let peer_manager_for_fulltext = coordinator.peer_manager.clone();
 
         // 6. Swap write path — cluster coordinator handles replica routing.
         self.write_path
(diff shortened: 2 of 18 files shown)


bkearns:
**Live multi-node validation: PASS** (the t_4ae47a9f close-out gate, run 2026-07-16 on fly.io).

Setup: 3 machines (shared-cpu-1x, **2 GiB cap** — the intentional assertion, never raised), image built from this branch, RF=3, 50k documents × 2 KB all matching the broad terms, real CQL wire protocol from an external driver through per-node proxies. `--restart no` so an OOM-killed node would stay down as loud evidence.

Load (515 s sustained, 12 concurrent workload threads across all three coordinators): 6 broad **no-LIMIT** `fts_match` scans returning all 50k rows each (the exact query shape that OOM-killed the 2 GiB-capped dev cluster — t_8fc24ce2), multi-term scans, 45 page-abandon churns (cancel path), 60 hostile `LIMIT 10` queries (the deleted-escalation shape).

Results:
- **Peak RSS: node1 132 MiB, node2 82 MiB, node3 79 MiB** — ~6.5% of the 2 GiB ceiling the legacy path blew through (sampled every ~3 s via each node's `ferrosa_process_resident_memory_bytes`, 42 samples).
- **Completeness under stress:** every broad scan returned exactly 50,000 rows on every coordinator; every LIMIT 10 returned exactly 10.
- All three machines alive at the end; machines destroyed after the run.

The pre-existing scaffold harness (deploy/fly-stream-scan) turned out to be non-executable beyond provisioning (`ferrosa-ctl seed`/`scan-*` don't exist; `CREATE CUSTOM INDEX` isn't parsed), so this run drove everything over the real client protocol instead — arguably a more faithful reproduction of the original incident. Provisioning notes worth folding back into the harness someday: fly private networking is IPv6-only, so `FERROSA_INTERNODE_BIND=[::]:17000` is required (default `0.0.0.0` silently forms no cluster), and the metrics listener binds guest-loopback so RSS must be sampled via ssh, not proxy.

PR #266, opened 2026-07-16
fix(cluster): resolve raft target host_id per RPC so late registration heals the network client

bkearns:
## Problem — why PR #264 (a docs-only PR) went red

`three_node_cluster_elects_raft_leader` failed on #264's CI run with the candidate stuck at **term=1 for 30 s** while both peers stayed Learners. #264 can't have caused it — the race is in main's formation code, and the same mechanism bit the live dev cluster on 2026-07-16 (an AppendEntries retry loop erroring `unknown peer: 00000000-…` every 500 ms for hours, never healing even after its peers came back — forge t_05acdc8a item 4).

**Root cause:** openraft caches one network client per target, and `FerrosRaftNetworkFactory::new_client` resolved the `node_id → host_id` mapping **once at construction**. A client built before `register_node` landed froze `Uuid::nil()` forever — openraft's backoff retries through the same frozen client, so the unreachability is permanent. In the formation race, the seed initializes raft before its peers' registrations land; its Vote RPCs are unroutable, nobody ever hears an election, and the 30 s deadline expires. The old test `factory_unknown_node_yields_nil` was *pinning the bug*.

## Fix

The client now stores the target node_id plus the shared registration map and resolves on **every RPC** (an uncontended `RwLock` read — noise next to a network round-trip). An unregistered target maps to `openraft::Unreachable`, so openraft backs off, and the retry succeeds as soon as `register_node` lands (formation loop or membership observer). A late registration heals the existing client instead of requiring a process restart.

## Verification

- New pin `late_registration_heals_existing_client`: a client created before registration must observe the mapping afterwards — fails against the old frozen-at-construction design.
- `factory_unknown_node_yields_nil` (which asserted the buggy nil bake-in) replaced by `factory_unknown_node_is_unresolved`.
- `cluster_formation` suite green locally, including the test that failed on #264's run; full ferrosa-cluster lib suite green (1059); fmt/clippy clean.
- A 12-iteration formation-race harness under 6-way CPU pressure: green on this branch (the race is timing-dependent and reproduces on starved CI runners; the mechanism is proven by the code + CI log + live incident).

After this merges, #264 should be rebased onto main so its checks run with the fix in — its own diff is docs-only and was never the problem.

### DIFF
--- a/ferrosa-cluster/src/raft/network.rs
+++ b/ferrosa-cluster/src/raft/network.rs
@@ -107,18 +107,29 @@ impl RaftNetworkFactory<FerrosRaftConfig> for FerrosRaftNetworkFactory {
     type Network = FerrosRaftNetwork;
 
     async fn new_client(&mut self, target: u64, _node: &BasicNode) -> FerrosRaftNetwork {
-        // Resolve the UUID from the sync map.
-        let target_host_id = self.resolve_host_id(target).unwrap_or_else(|| {
-            // openraft does not allow returning an error here.  Return a
-            // nil UUID; the first RPC will fail with Unreachable which will
-            // trigger openraft's backoff logic.
-            tracing::error!(node_id = target, "no host_id registered for Raft node");
-            Uuid::nil()
-        });
+        // The host UUID is resolved PER RPC, never baked in here: openraft
+        // caches this client per target, and a client built before the
+        // node_id → host_id registration landed used to freeze `Uuid::nil()`
+        // forever — every retry then failed with "unknown peer: 00000000-…"
+        // and the cluster never healed even after the peer registered. Two
+        // live manifestations: the CI formation race (seed initializes raft
+        // before `register_node` runs for its peers → Vote RPCs unroutable →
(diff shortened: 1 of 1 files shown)


bkearns:
Scope honesty note from pre-merge stress testing: a 12-iteration run of the formation suite under 6-way CPU pressure still produced **1 failure on this branch** — so this PR fixes one proven, permanent-failure mechanism (the frozen-nil client, which matches the live 2026-07-16 incident exactly and is pinned by the new unit test), but it is NOT the whole formation race.

Evidence from the surviving failure (differs from the frozen-nil signature): node1=Candidate term=1 (never retries the election), node2=Learner term=1 (it DID receive node1's vote traffic — transport was working), node3=Learner term=0 (never contacted). A candidate that fails its first election round should re-elect with term=2 after the election timeout; staying at term=1 for 30 s means the election retry never fires. That is a separate defect — likely in election scheduling/suppression, not RPC routing — and is now tracked with this evidence on forge t_05acdc8a. This PR stands on its own merits; the remaining race needs its own investigation.

PR #308, opened 2026-07-26
Materialization audit, Elle cert hardening, formation election-stall fix, replayable Accord log

bkearns:
Scope grew substantially since opening — this PR now carries four related workstreams, rebased onto e6c768cd:

## 1. Materialization audit coverage + replayable Accord protocol log (original scope)
Exhaustive-by-default materialization lint (NON_SERVING_CRATES + unclassified-crate findings), graph truncation fix (capless `collect_to_graph_result` — a mid-drain cap silently truncates with no partial-result flag; bounds belong upstream in the stream), framed Accord protocol log + crash recovery.

## 2. Elle certification hardening (6 commits)
The 2026-07-26 nemesis cert was **vacuous**: the generator pinned all workers to one node, so killing the coordinator stopped 100% of traffic (1026 :ok / 13369 :fail, `valid? true` over ~1000 single-coordinator appends). Fixes: workers spread across nodes (each pinned to its node, preserving the one-txn-one-connection invariant), dead sessions replaced on any failure, rejoin latency measured properly, workload sized to dominate faults, coordinator-restart fault added, + a local podman certification harness (`deploy/local-elle/`) so harness breakage is caught before a cloud run.

## 3. Formation diagnosability (2 commits)
The formation test installed no tracing subscriber (CI failures printed nothing usable), and `RUST_LOG=openraft=debug` HIDES the formation race (20/20 pass with it) — so the durable diagnostic is a metrics-watch timeline recorder (25ms sampling, shared epoch, voters/learners dump on failure).

## 4. Formation election-stall fix (2 commits: red test + fix) — the bug that failed this PR's own Test + Coverage
Root cause (t_b0aac0d3): `raft_enable_pre_vote` defaulted true, but the pinned openraft fork hard-gates the tick election path behind a pre-vote round whose transport `FerrosRaftNetwork` never implements — every probe is an automatic NO, so after the seed's single `initialize()`-driven election at term 1, no election can ever fire again. A transient first-round vote loss became a permanent stall (~3% under CPU starvation; CI job 89743146291 was exactly this).
Fix: default the gate OFF until the transport lands (t_32cb5ad3); tick path then re-elects and self-heals, like upstream. Regression guard: `candidate_re_campaigns_while_peers_are_down` (term must advance past 1 with peers down; deterministic red on the unfixed tree).
**Verification: 250 consecutive CPU-starved formation runs, zero failures** (two runs of 200+50, both externally killed mid-run by machine pressure, zero failures in every completed iteration), vs pristine main's measured ~3% (1/30 control + 5/85 across instrumented runs).

Spec: `specs/implemented/bug-cluster-formation-pre-vote-election-stall.md`. FMEA updated: CL-2 no longer claims PreVote as a live mitigation (the election_guard watchdog is load-bearing); new CL-15 row.

Pre-existing red-check note for reviewers: Test + Coverage may still hit two OTHER catalogued load-sensitive families unrelated to this diff — compaction bounded-wait (t_283f37df) and fair_admit share-capture (t_ffb53cf9). Attribute against those before assuming this PR.


bkearns:
Reviewer coverage statement (honest scope, per-commit):

**Deep line review (by me, this cycle):**
- `6e7a2750` red test + `60af610c` pre-vote gate fix — verified the RED commit contains zero config changes (test genuinely fails at that commit), the fix's mechanism against the fork source (raft_core.rs:1535-1546 gate, network.rs missing override, engine_impl.rs:215 initialize-elect bypass), the inverted ADR-012 assert's rationale, and the FMEA honesty fix (CL-2 no longer credits PreVote; guard reclassified load-bearing). Verified 250/250 CPU-starved formation runs clean vs main's ~3%.
- `13be72ee` graph rebase repair — both new #306 call sites passed `usize::MAX` (= no cap), so dropping the argument is semantics-preserving; verified against main's 4-arg silent-truncate version.
- `049025d5` diagnostics — shared-epoch timeline recorder (a per-node-epoch version produced a false causal paradox during investigation; the shared epoch is the fix), voters/learners in the failure dump.
- `07f7c145`/`274261f5` local-elle harness — reviewed script; 14.7MB run artifact gitignored, not committed.

**Reviewed as a set (Elle hardening, 6 commits `b8287255..405c22d3`):** worker-spread preserves per-worker single-target pinning under FallthroughRetryPolicy (the one-txn-one-connection invariant); dead-session replacement + rejoin measurement address the t_0055fc1e vacuous-cert findings. The nemesis certification MUST be re-run post-merge before any "certified under faults" claim is repeated — tracked in t_08d33ff8 (breakdown criterion: generator outlives faults, :ok/:fail ratio healthy, not just `valid? true`).

**Pre-existing from original PR scope (not re-reviewed this cycle):** the materialization-audit/xtask, Accord framed-log/crash-recovery, and postgres offload commits predate this convergence effort and rode through the rebase unchanged (3-way, no content conflicts outside the graph repair noted above).

bkearns:
Restructured per owner review: the two blocking incomplete features are SPLIT OUT — graph cap-removal (finding 2) now lives on feat/graph-capless-collect (stacked on this branch; lands with/after the streaming-executor work, t_3f2f961a), and the Accord framed-log + orphan-resolution commits (finding 3) on feat/accord-framed-log-recovery (returns when migration + startup replay wiring + kill/restart test are complete, t_6691c002). #308 is now 16 commits: materialization-lint coverage + triage docs, postgres executor offload, Elle certification hardening, formation diagnosability, and the pre-vote election-stall fix. The p0-oom-audit passes in enforce mode on the slimmed branch (the result-bound rule change rides with its graph fix on the split branch). Nemesis-effectiveness finding (5) tracked as t_2a37f1bf against the certification harness — the harness commits remain here, the fatality hardening lands as a follow-up before the next certification claim. Local gates re-run: clippy 0, cargo doc -D warnings 0, formation suite 4/4.

bkearns:
Rebased onto 1ecaa060 (post #312/#315). The merge queue marked this UNMERGEABLE at position 3 because #312 and this PR both edit ferrosa-cluster/specs/fmea.md — #312 added the N-way-merge cause (b) + test reference to CL-14, this PR adds the CL-15 pre-vote row. Resolved by keeping #312's CL-14 verbatim and appending CL-15; both changes survive (verified: the nway test reference and the pre-vote row are both present). Re-gated after rebase: fmt clean, clippy 0, p0-oom-audit enforce-mode 0 findings, ferrosa-cluster lib 1089/1089.

PR #313, opened 2026-07-28
docs(storage): O_DIRECT compaction I/O design (salvaged from Sprint 0 worktree)

bkearns:
## Summary

Salvages the design docs for an **opt-in, default-off, Linux-only `O_DIRECT` read path for compaction and index-rebuild scans** (task `t_bed7ed0f`, from the PR #279 I/O page-cache audit). Docs-only — no source touched.

Lands four `status: proposed` docs under `specs/proposed/o-direct-compaction-io/`:

- `decisions.md` — ADR: opt-in Linux `O_DIRECT` for compaction/rebuild I/O
- `architecture.md` — `IoMode` seam + `DirectReadAt` + audited Linux adapter, with current/proposed Mermaid flows
- `fmea.md` — failure-mode analysis (top RPN: mixed direct/buffered access, dropped unaligned tail, partial-data checksum)
- `project-plan.md` — five workstreams; **Sprint 0 is a measurement GATE that may cancel the project**

The design is read-side only (writers stay buffered), measurement-gated (default-off until a Linux/NVMe cgroup A/B proves page-cache eviction materially hurts the foreground hot set), and isolates all `unsafe`/platform code in one audited adapter.

## Flag for reviewer: relationship to existing O_DIRECT results on `main`

These are **design docs and state no benchmark results**, so there is no direct results-contradiction to correct. But there is a scope/evidence relationship worth calling out, since `deploy/fly-bench/ODIRECT-*-RESULT.txt` already exist on `main`:

- **Different feature/scope.** This design is a read-side compaction/rebuild **reader** (`DirectReadAt`, env `FERROSA_COMPACTION_IO_DIRECT`, task `t_bed7ed0f`). The results on `main` are for an already-built write-side **`DirectWriter`** (`FERROSA_SSTABLE_DIRECT_IO`, task `t_d1af57de`). Same syscall, different code path.
- **Existing evidence is relevant to this design's Sprint 0 gate.** The write-side A/B (`ODIRECT-AB-HARSH-RESULT`, `ODIRECT-RAMP-RESULT`) found **no demonstrated O_DIRECT benefit on fly fast NVMe** — the device never saturated (~8% util) and the writeback-freeze did not reproduce. Critically, `ODIRECT-RAMP-RESULT` notes the **cache-preservation hypothesis is UNMEASURED** (concurrent full-table scans never completed within the ramp), which is exactly the hypothesis this read-side design's Sprint 0 gate exists to test.
- **Not a contradiction, a caveat.** The design's motivating claim (compaction scans evict the foreground hot set under a bounded cgroup) is framed internally as unproven and gated behind Sprint 0. Existing results tested a different regime (4GB RAM / 64MB cache / no cache-pressure, fast NVMe) than the design's target (2 GiB cgroup cache-pressure). Sprint 0 should incorporate the existing write-side results and use bounded/token-range scans so the read-side cache hypothesis is actually measured before any code is written.

No edits were made to the salvaged docs; this flag is surfaced here rather than in-file per the salvage instructions.

## Verification

Docs-only; `git diff --stat origin/main` shows only files under `specs/proposed/o-direct-compaction-io/` (4 files, +447). No cargo gates needed.

### DIFF
--- a/specs/proposed/o-direct-compaction-io/architecture.md
+++ b/specs/proposed/o-direct-compaction-io/architecture.md
@@ -1,0 +1,184 @@
+---
+title: "Opt-in Linux O_DIRECT compaction/rebuild reader — architecture"
+status: proposed
+component: storage-io
+task: t_bed7ed0f
+executive_summary: >
+  Extends ferrosa-sstable's ReadAt seam with a default-off, Linux-only direct-I/O
+  reader for compaction and index-rebuild scans. A safe IoMode selects between the
+  existing buffered FileReadAt and a new DirectReadAt whose unsafe/platform code
+  (O_DIRECT open, aligned allocation, raw pread) is isolated in one audited Linux
+  adapter. DirectReadAt probes STATX_DIOALIGN, reads aligned supersets into bounded
+  aligned bounce buffers, copies out the requested slice (handling unaligned
+  head/tail), verifies checksums on reassembled data, and falls back to buffered on
+  any unproven alignment or unsupported filesystem/platform. Compaction/rebuild use
+  a SEPARATE reader pool; foreground reads and the shared foreground pool are
+  untouched. Instrumentation captures buffered/direct bytes, fallbacks, tail
+  fallbacks, foreground latency, faults/refaults, PSI, and device latency to feed
+  the measurement promotion gate.
+last_revised: 2026-07-21
+---
+
+# Opt-in Linux `O_DIRECT` compaction/rebuild reader — architecture
+
+## Problem
+
+Compaction and index rebuild stream cold `Data.db` bodies through the OS page cache
+via buffered `FileReadAt` (pooled in `compaction/executor.rs` as
+`SharedReaderPool<FileReadAt>`). Under a bounded cgroup this evicts the foreground
+random-read hot set, raising serving tail latency and refaults.
+
+## Goal
+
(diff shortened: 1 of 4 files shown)


bkearns:
Reviewed: docs-only confirmed (4 files under specs/proposed/o-direct-compaction-io/, +447, zero source; grepped for results-style claims — none, it is status:proposed design). The scope flag in the PR body is correct and worth keeping visible: this is the READ-side compaction reader design (FERROSA_COMPACTION_IO_DIRECT, t_bed7ed0f), distinct from the merged WRITE-side DirectWriter results in deploy/fly-bench — and those existing A/Bs found no demonstrated write-side O_DIRECT benefit with the cache-preservation hypothesis explicitly unmeasured, which is precisely what this design's Sprint 0 gate proposes to measure. Recommend Sprint 0 folds in those results and uses bounded/token-range scans, per the flag.

bkearns:
Test + Coverage failure attributed: fair_admit::tests::tenants_get_equal_share_regardless_of_query_count (ferrosa-sched) — ratio 2.26 vs the 2.0 bound. This PR's diff is specs-only, so the failure is pre-existing on main; filed as forge t_ffb53cf9 (third distinct timing-sensitive family on main, alongside t_283f37df and the formation race). The fairness bound miss is a real signal for the scheduler epic — do not widen the bound. This PR is mergeable on content; the red check belongs to main.

PR #314, opened 2026-07-28
fix(storage): preserve cell ordinals for unflushed pre-ALTER rows

bkearns:
## Problem

Row cells are stored as `(u16 column_ordinal, CellValue)` where the ordinal is the column's position in `regular_columns` **at write time**. Regular columns are ordered by the column-name comparator, so `ALTER TABLE ADD <col>` of a column that sorts before existing columns re-numbers every later column's ordinal.

ST-16 (df5152bf) fixed the index-declaration side of this. The remaining crash window: rows written under the **old** schema that are still dirty in the memtable/replay at ALTER time were flushed under the **new** schema — their SSTable `SerializationHeader` then claimed the new layout while the cells carried old ordinals, so reads misattributed values to the wrong columns (`name` reads back NULL, the new column reads back `'John'`), and index backfills over legacy SSTables built postings from the wrong cell.

## Fix

- `StorageEngine::update_table_schema` now flushes dirty pre-ALTER rows under the **old** schema before swapping layouts, while holding the table-map write lock. Engine writes need the table-map read lock before commit-log append / memtable insert, so no row can land between the flush barrier and the new schema. The flushed SSTable's `SerializationHeader` preserves the write-time layout; post-ALTER reads remap physical ordinals by column name (existing header-remap path). The covered commit-log position is discarded and `schema.json` persisted after the swap.
- Index backfill (`add_index*` and the eager post-flush path) maps the current regular-column ordinal and filtered-index predicate clauses through each SSTable's stored `SerializationHeader` before scheduling the build job (`TableStore::source_regular_ordinal_for_sstable`, `source_filter_predicate_for_sstable`). A column absent from a legacy SSTable is marked indexed with no postings for that generation instead of probing a different old column at the same ordinal.
- The flush barrier composes with the flush parallelization on main: the parallel encode lives inside `TableStore::flush` (flush_executor pool), so nothing that main parallelized is serialized by the barrier.
- Spec: `specs/todo/pre-alter-rows-stale-cell-ordinals.md` moved to `specs/implemented/` with the resolution; ST-16 FMEA row (RPN 224 → 32) and `ferrosa-storage` README/overview updated.

## Tests

- `engine::tests::update_table_schema_preserves_unflushed_pre_alter_row_ordinals` — writes under the old layout, ALTERs so a new column sorts first, asserts the pre-ALTER row's cells remap to the current ordinals (and are not misattributed to the new column), both live and across a restart/replay boundary.
- `store::tests::index_backfill_maps_current_ordinals_to_legacy_sstable_source_ordinals` — current ordinals 1/2 backfill from legacy physical ordinals 0/1; a newly added column returns `None`.
- `store::tests::index_backfill_remaps_filtered_predicate_to_legacy_sstable_source_ordinals` — predicate clauses remap to legacy ordinals; a predicate on a column absent from the legacy SSTable becomes an empty conjunction that evaluates false.

`cargo fmt --check`, `cargo clippy --all-targets -p ferrosa-storage` (zero warnings), and the crate suite pass apart from the pre-existing failure below.

## Pre-existing failure note

`engine::tests::compaction_output_uploaded_to_s3` failed once in the full-crate run on this branch (empty manifest after the bounded wait). The same test — along with 6 others in the compaction/manifest bounded-wait family — fails on a **pristine origin/main** worktree under host load, so it is not attributable to this change. Tracked as forge task t_283f37df.

## Context

ST-16 follow-on; forge t_4a1afc2d. Residual from the ST-16 FMEA row ("rows written BEFORE the ALTER still carry old cell ordinals") is closed by this PR.


bkearns:
Reviewed in depth (both halves verified against main's actual code, not just the diff).

**What's right (verified):**
- Empty-conjunction sentinel is sound: all three evaluators in `ferrosa-index/src/filtered.rs` (:43, :65, :137) begin `!predicate.clauses.is_empty() &&`, so remapping an unsatisfiable predicate to an empty conjunction genuinely produces a no-entry build, not an index-everything.
- Backfill None-handling is correct: a column absent from a legacy SSTable → `mark_indexed` + skip is completion, not a silent skip (nothing to index), and avoids permanently-pending trackers.
- Flush-under-write-lock keeps main's parallel encode (the rayon pool lives *inside* `TableStore::flush`) — the barrier serializes admission, not the encode.
- `mark_pending` 5-arg / `mark_indexed` 4-arg carried correctly across the post-base signature change.

**Required change 1 — the barrier comment overclaims.** It states writers need the table-map read lock "before they can append to the commit log or enter TableStore::write". Not so: main's write path (`engine.rs:5767`) appends to the commit log *before* taking the read lock. Surviving interleaving: writer appends M (old-ordinal cells) → blocks on the ALTER's write lock → ALTER flushes + swaps + releases → writer inserts M's old-ordinal row into the new-schema memtable; M also replays mis-tagged after a crash (its cl_pos postdates the discard capture). The fix closes the dominant window (all buffered rows) — but the comment must describe the residual racing-writer window honestly, and the spec's Resolution section should carry it as a known residual. Follow-up task will track closing it (schema-epoch validation at memtable admission or moving append under the guard).

**Required change 2 — silent identity fallback.** `column_mapping_for_sstable` returns `None` on `open_reader(desc).ok()?`, and `source_regular_ordinal_for_sstable` then falls back to the *identity* mapping. For a legacy reordered SSTable whose reader hits a transient IO error, that reintroduces exactly the wrong-ordinal probe this PR exists to kill — silently. Wants a `warn!` distinguishing "SSTable not in view (vanished, benign)" from "reader open failed → identity fallback", per the fail-loud rule. Same applies to `source_filter_predicate_for_sstable`'s clone-through path.

**Noted, no action:** the single full-suite failure (`compaction_output_uploaded_to_s3`) is attributed to the pre-existing load-dependent family on pristine main (7/1020 failed on an untouched control worktree) — t_283f37df. Correctly cited in the PR body.

Verdict: approve once the two required changes land; neither alters the fix's structure.

bkearns:
Follow-up task for the racing-writer residual is filed: forge t_237efb08 (schema-epoch validation at memtable admission / append-under-lock / name-keyed cells — directions enumerated there). Reference it from the barrier comment and spec residual.

bkearns:
Both required changes landed in afa7d43d:

1. Barrier comment reworded — states the write-lock hold excludes memtable admission + the `last_commit_log_position` store, and names the racing-writer residual (commit-log append happens before the read lock; blocked writer lands an old-ordinal row post-swap and replays mis-tagged), referencing t_237efb08. Spec Resolution carries the same residual.
2. `column_mapping_for_sstable` WARN-logs on `open_reader` failure before degrading to current-schema ordinals; the benign not-in-view case stays silent.

Verified: `cargo fmt --check` clean, clippy zero warnings, all three new tests pass.

bkearns:
Follow-up afa7d43d reviewed: both required changes are correct and complete — the barrier comment/spec now state exactly what the write-lock hold excludes and carry the racing-writer residual with t_237efb08, and the reader-open failure path is warn-logged with the benign vanish case correctly left silent. **Approved.** (Note for merger: CI's Test + Coverage may hit the pre-existing t_283f37df compaction family or the formation race — attribute against those tasks before assuming this diff.)

bkearns:
Owner-review response, part 1 of 2 (fail-open finding): FIXED — three-state mapping resolution; unreadable header now fails CREATE INDEX loudly (backfill) or logs+leaves-tracker-pending (eager), never guesses a layout. 7 affected tests green, clippy 0.

Part 2 (racing-writer window) analysis: moving the commit-log append under the tables read lock is NOT sufficient — the stale ordinal tagging attaches when the row is BUILT upstream, so a writer that built pre-ALTER and blocked on the lock still inserts stale-tagged data post-swap. Sound closure = schema-epoch validation (row carries build epoch; validated under the write guard; mismatch fails loud, caller rebuilds). Local-node epoch is bounded work; forwarded cluster writes raise a distributed schema-version question (coordinator-built rows validated against replica epochs) that belongs with the raft-DDL schema-agreement line. Design + implementation tracked on t_237efb08; this PR stays unmerged until that closure lands per the review.

PR #315, opened 2026-07-28
fix(storage): preserve compaction inputs until the pending-upload marker is durable (#235)

### LINKED ISSUE #235: P0: typed_edges visibility divergence after OOM/restart; diagnose compaction/SSTable manifest corruption
## Summary

After the local `fmem-dev` cluster soaked overnight, `node1` and `node2` were OOM-killed at `2026-06-30 07:10:40-07:10:41 PDT`. After restart, graph-owned data is read-visible inconsistently across coordinators, especially `agent_memory.typed_edges`. This may be a compaction/SSTable manifest/load issue rather than intentional deletion, but it is P0 until proven otherwise.

This issue is for a TDD diagnostic/fix path. Do not change product semantics to cap reads or silently drop rows. The desired behavior is streaming reads with stable live-row visibility and fail-loud corruption handling.

## Evidence from local cluster

Cluster: podman compose project `fmem-dev`.

Containers:
- `fmem-dev_node1_1` CQL `19042`: `RestartCount=1`, `OOMKilled=true`, restarted `2026-06-30 07:10:40 PDT`
- `fmem-dev_node2_1` CQL `19043`: `RestartCount=1`, `OOMKilled=true`, restarted `2026-06-30 07:10:41 PDT`
- `fmem-dev_node3_1` CQL `19044`: no restart/OOM in this window

Observed symptoms:
- Entities are not obviously below the expected scale: `agent_memory.entity_store=78,207`, `document_chunks=69,867`.
- `typed_edges` visibility diverges by coordinator. Key-only scans returned different live keysets:
  - one pass: `node1=10,488`, `node2=10,993`, `node3=13,744`
  - repeat pass: `node1=10,993`, `node2=10,993`, `node3=13,355`
- The differing keys are mostly nil-session graph edges: `contains`, `references`, `calls`, `depends_on`.
- `COUNT(*)` on `typed_edges` is unstable and disagrees with key scans:
  - earlier: `node1/node2=46,066`, `node3=50,807`
  - later: `node1/node2=50,807`, `node3=46,066`
  - key scan at the same scale returns only ~11k-13k unique primary keys.
- No evidence of user/requested deletion in quick checks: `forget_journal=0`, `retraction=0`.

Startup/storage log indicators:
- node1 startup loaded only `13` local `agent_memory.typed_edges` SSTables and logged missing pending-upload typed-edge generations.
- node2 startup loaded only `8` local `agent_memory.typed_edges` SSTables.
- node1/node2 startup logged unresolved `idx_typed_edges_dst` / `idx_typed_edges_edge_type` target columns during the same load window.
- Both restarted nodes emitted many `required SSTable component absent in S3 and local disk — skipping stale manifest entry` messages across tables.

Manifest/S3 inventory for `agent_memory.typed_edges`:
- node1: manifest entries `82`, local Data generations `13`, S3 Data generations `54`, manifest-not-S3 `72`
- node2: manifest entries `28`, local Data generations `8`, S3 Data generations `32`, manifest-not-S3 `21`
- node3: manifest entries `83`, local Data generations `11`, S3 Data generations `36`, manifest-not-S3 `73`

A forensic copy is being taken at:
`/Users/bkearns/data/ferrosa-memory-forensics/20260630T144820Z/`

## TDD test list

Start by writing failing tests that reproduce the safety properties before implementing fixes.

- [ ] Restart/load fails loud, or reports degraded table state, when a manifest references a generation whose required `Data.db` component is absent locally and in S3.
- [ ] Compaction manifest update is atomic: no manifest generation is made authoritative until every required component is durably promoted and uploaded/recorded.
- [ ] Compaction cleanup never removes the last readable copy of a generation still referenced by the authoritative manifest.
- [ ] Startup reconciliation drops or marks stale manifest entries consistently without making `COUNT(*)` include unreadable rows.
- [ ] `SELECT COUNT(*)` and a full streaming primary-key scan agree on live-row cardinality for a table with stale/compacted SSTable generations.
- [ ] After simulated OOM/crash during compaction or S3 sync, restart either recovers the same keyset from local/S3 or fails loudly with a diagnostic error; it must not silently serve a partial keyset.
- [ ] Secondary-index registration/rebuild is ordered after table schema registration so `idx_typed_edges_*` does not transiently become silently unusable after restart.
- [ ] Multi-coordinator read consistency test: after restart of two nodes, key-only scans of the same table at CL=ONE across all coordinators converge or surface an explicit degraded/inconsistent state.
- [ ] Regression test for graph-owned backing tables, at minimum `agent_memory.typed_edges` and `system_graph_agent_memory.adjacency`.

## Diagnostic acceptance criteria

- The failing test proves whether `COUNT(*)` is counting stale manifest/SSTable records or live rows.
- The failing test proves whether compaction manifest updates can reference missing components.
- The fix preserves move/streaming semantics and does not introduce any silent max-row cap.
- Corruption/self-heal paths must fail loud with table/generation/component details when data cannot be safely served.
- No repair/delete/cleanup behavior should be part of the first fix unless separately tested and explicitly reviewed.

## Operational caution

Do not run cleanup, repair, compaction, volume deletion, or data restore against the local cluster until the forensic copy is complete and the recovery path is agreed. Current evidence suggests read-visible divergence; permanent data loss is not yet proven.

bkearns:
## Summary

Additional durability hardening for #235, beyond the eviction/upload fix merged in #236. This does **not** close #235's visibility-divergence investigation — it removes one more data-loss window on the compaction path.

Previously, `poll_compactions` reclaimed the compaction inputs **before** the output's pending-upload record was written, and a failed pending-log write/open was only a `warn!` — inputs were deleted anyway. Until that marker is fsynced, the output is not recoverable after a crash (nothing replays its upload), so a crash or disk-pressure eviction in that window strands the manifest with no readable copy of the data.

## Changes

- `ferrosa-storage/src/engine.rs` (`poll_compactions`):
  - Input cleanup is hoisted into a `cleanup_inputs()` closure invoked only on paths where the output is already safe: pinned tables (NVMe-only, no upload), no upload manager / no store configured, and — on the S3 path — **only after** `PendingUploadsLog::add_compaction_entry` succeeds.
  - A pending-log open or append failure now logs `error!` and `continue`s, **preserving every input** (fail loud, never delete on an unrecorded output). Anomalous paths (unparseable output id, zero-byte output) also preserve inputs.
  - `pending_log` is now held by value from the successful open, removing the old `Result`-guarded access at entry removal.
- `ferrosa-storage/specs/fmea.md`: ST-12 mitigation updated to cover the compaction-side ordering and the new test.

## Testing

- New test `compaction_keeps_inputs_when_pending_upload_log_cannot_be_written`: replaces `pending-uploads.log` with a directory so the marker write fails, then asserts every input `Data.db` survives multiple `poll_compactions` iterations. Verified failing against the pre-fix ordering (inputs deleted) and passing with the fix.
- `cargo fmt` and `cargo clippy --all-targets -p ferrosa-storage` clean.
- Full `cargo test -p ferrosa-storage`: 1021 lib tests + all integration suites pass, 0 failed, 0 ignored.

Refs #235.


bkearns:
Reviewed (diff + ordering verified against main's poll_compactions):

**Correct:**
- `cleanup_inputs` closure invoked on exactly the safe paths: pinned-table, no-upload-manager, no-store, and post-`add_compaction_entry`-success. The unconditional up-front eviction is gone.
- Both pending-log failure arms (`open` and `add_compaction_entry`) are `error!` + `continue` with inputs preserved — fail-loud, and crash-safe: the manifest still references the preserved inputs (C2.2 keeps manifest updates behind upload confirmation), so a crash recompacts rather than losing data.
- `remove_entry` warn-nonfatal is justified by idempotent replay.
- Red/green proof (eager cleanup temporarily re-added → test fails) demonstrates the guard is load-bearing against current main.

**Requested change (test vacuity):** `compaction_keeps_inputs_when_pending_upload_log_cannot_be_written` asserts inputs exist after 40×50ms of polling. If compaction never completes inside the window (CI starvation — see t_283f37df for how real that is on this exact family), the assert passes vacuously. Add a completion witness before the input assert — e.g. assert a compaction output generation appeared under the table dir (or the compaction counter advanced) so the test fails loudly as "compaction never ran" instead of green-by-timeout. Keeps it un-invertible: today it can only false-pass, never false-fail; after the guard it can do neither.

Datapoint for t_283f37df: this branch's full ferrosa-storage run was 1021 passed / 0 failed under lighter concurrent load — consistent with the load-scaling attribution.

Verdict: approve once the completion witness lands.

bkearns:
Follow-up bbdcd9d3 reviewed: the completion witness is sound — output promotion (fresh generation, pre-pending-log) is observed before the preservation assert, so the test can neither false-pass on a stalled compaction nor false-fail (witness verified to still hold under the red emulation). **Approved.** Same merger note as #314: any Test + Coverage failure here should be attributed against t_283f37df / the formation race before assuming this diff.

PR #316, opened 2026-07-28
feat(accord): read-in-transaction — buffer SELECTs, return rows atomically at COMMIT (slice 1/1a)

bkearns:
Read-in-transaction slice 1/1a, salvaged from the pre-#285 local branch and ported onto the merged TransactionRegistry model (epic t_5a8d67c9 lineage).

**Storage seam**: `TransactionRead` / `CommitReads` / `TransactionCommitter::commit_with_reads` — positional read-set evaluated at the transaction's agreed commit timestamp. Default impl commits writes and accepts only an EMPTY read-set; a non-empty read-set against a committer without read-in-transaction support is a loud `CommitError` (silently returning defaults would fabricate row-absence, since `None` means "absent at commit-t").

**CQL wiring** (ported onto post-#285 code): reads buffer in `CqlTransaction.reads` inside the registry's `TransactionEntry` — txn-id-owned, reaped by the same 10s deadline as writes; `stage_read` fails loud outside a txn and counts against the shared statement cap; `commit_transaction` routes through `commit_with_reads`. The SELECT-in-transaction router arm and the engine's multi-key read-at-commit-t are later slices — no production path stages reads yet.

Port review note: the original slice-1a test pinned the default impl's silent-drop behavior; that test was authored in this same branch and was rewritten (not weakened) to pin the strengthened contract from both sides — empty read-set commits, non-empty read-set errors loudly.

Spec: `specs/proposed/read-in-accord-transaction.md`.

### DIFF
--- a/ferrosa-cql/src/router.rs
+++ b/ferrosa-cql/src/router.rs
@@ -1861,8 +1861,13 @@ async fn commit_transaction(
         }
     };
     match txn.commit(committer.as_ref()).await {
-        Ok(CommitOutcome::Committed) => Ok(RouteResult::Result(crate::result::encode_void())),
-        Ok(CommitOutcome::Aborted { reason }) => {
+        // A write-only transaction acks with void. Read-in-transaction (returning
+        // the buffered SELECT's rows from COMMIT) is wired in the
+        // SELECT-in-transaction slice; a write-only COMMIT never carries reads.
+        Ok((CommitOutcome::Committed, _reads)) => {
+            Ok(RouteResult::Result(crate::result::encode_void()))
+        }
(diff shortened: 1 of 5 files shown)


bkearns:
Transparency note: the porting agent died (spend limit) after the rebase + clippy/fmt; I completed tests, found and fixed one fail-loud gap in its port (the default commit_with_reads silently dropping staged reads — now a loud CommitError, with the same-branch test rewritten to pin both sides of the strengthened contract), and pushed. So this PR is part-agent, part-reviewer work: the storage seam and CQL wiring got my full line review in the process; treat my involvement as reviewer-approved with the caveat that finisher and reviewer are the same party here — worth a second human glance at the commit_with_reads contract decision.

PR #340, opened 2026-08-18
fix(repair): stop calling a missing row a timestamp tie

bkearns:
A replica that is simply **missing rows** was classified as an unresolvable conflict, so repair streamed nothing and the replica stayed short forever while reporting success.

This is the operative cause of the persistent divergence on the host cluster — the Merkle XOR flaw in #339 is real but did **not** explain it.

## The bug

Ties are decided on the partition's **max** timestamp. That maximum says nothing about rows below it: a replica can be missing whole rows while its newest surviving cell shares a timestamp with the other side. The partition hashes differ, the timestamps compare equal, and the whole partition is written off as a tie — *"last-write-wins with equal timestamps is undefined"* — when there is nothing undefined about it. Every row the thin side holds is present and identical on the fat side. The union is the only sensible answer.

Two genuinely different situations were conflated:

| situation | ambiguity? | now |
|---|---|---|
| one side's rows are a **subset**, shared rows agree | none | stream the union |
| each side has rows the other lacks, shared rows agree | none | stream both ways |
| a row on **both** sides, same timestamp, different content | real | still a tie, still not streamed |
| partition deletion / static row differs | not row data | still a tie |

The third row is what the tie was introduced for, and it is deliberately unchanged — picking a side silently would discard someone's write.

## Evidence

On the host 3-node cluster:

```
distinct entity ids:  node1=78494   node2=78499   node3=78499
missing from node1: 5      unique to node1: 0
repair (all 3 coordinators): streamed_in 0, streamed_out 0,
                             sessions_ok 2/2, errors [], timestamp_ties 7
```

Strictly one-directional — node1 has nothing the others lack. Repair was **finding** the divergence on every run (the tie counter is never zero) and declining to act on it.

Ruled out first: a measurement race (reverse read order gives identical counts; same node twice 20s apart gives delta 0), tombstones and conflicting writes (those leave rows unique to *each* node), and the Merkle XOR cancellation (all the missing rows have distinct content hashes, and cancellation needs equal hashes cancelling in pairs).

## TDD

The subset test was written first and confirmed failing against the old classifier, which recorded the partition as a tie and streamed nothing. Also pinned: the both-ways case, the genuine-conflict case (must remain a tie), and the partition-deletion case.

## Verification

1095 `ferrosa-cluster` tests pass. `fmt --check`, `clippy -D warnings` and `cargo doc` all clean locally before pushing — the doc gate caught a public-doc link to a private item, which is exactly the check that has failed PRs here before.

## Relationship to #339

Independent, and both worth landing. #339 fixes a latent detector flaw (equal content hashes cancelling in a Merkle leaf); this fixes the classifier that was discarding real, detected divergence. #339 alone does not repair this cluster — verified by deploying it and re-running repair, which still reported 0 streamed.

Not merging this myself: standing rule is no auto-merge on public repos.

---

# Update: reconciliation moved from the partition to the row

## Correction to the original commit

The first commit fixed only `diff_partition_sets`. **`executor.rs` calls `diff_partition_sets_streaming`**, so on a live cluster it changed nothing. My earlier claim that the fix was applied to both paths was wrong: `classify_equal_timestamp` had exactly one call site, and the streaming path emitted `Tie` unconditionally.

The improvement measured on the cluster after deploying this branch alongside #339 (`partitions_streamed_in: 2`, ties 7 → 1) came from **#339** — the Merkle fix changed which leaves are seen as divergent — not from this PR's classifier.

## Are the remaining ties genuine conflicts?

No. Inspecting the one partition still reported as a tie, `(tenant 9a5f8fbf, session 32ee0a9d)`:

```
node1 = 4 rows     node2 = 7     node3 = 7

3 rows  absent from node1 entirely  — node2 and node3 byte-identical
4 rows  present on node1 as EMPTY SHELLS:
        dd11fed6   node1: 0 populated fields   node2: 9
        ddf9e5c6   node1: 0 populated fields   node2: 9
        a64ce319   node1: 0 populated fields   node2: 8
        ec00ea62   node1: 0 populated fields   node2: 8
```

node1 holds the primary-key marker with every non-key column `None`. A conflict requires two plausible versions; node1's side contributes nothing, and taking the populated copy cannot discard a write that was never there.

The 3 unambiguously-missing rows were trapped because classification was **per partition** — any one contested row condemned every neighbour. Since `entity_id` is a clustering column, one partition holds thousands of unrelated entities.

## What changed

Reconciliation now runs per row, in **both** diff paths:

| Case | Verdict |
|---|---|
| Row on one side only | stream it — no ambiguity |
| Both sides, one newer (row timestamp) | newer wins |
| Both sides, equal ts, cells a strict subset | superset wins — discards nothing |
| Both sides, same cell + same ts + different value | **conflict**, recorded in `timestamp_ties`, left alone |

Also removed the partition-wide max-timestamp comparison, which was the same bug pointing the other way: one recent row made a whole partition "newer" and swallowed every row only the other side held, while repair reported success.

Cells are keyed by `(column, path)` to match `ferrosa_storage::merge::merge_rows`; keying by column alone would treat two elements of a collection as competing writes. Only the rows that need to move are streamed — less write amplification than resending the partition, and the apply path already writes row-by-row.

## TDD

Four tests written first, all confirmed failing:

- `a_contested_row_does_not_block_the_rest_of_its_partition`
- `an_empty_shell_loses_to_the_populated_copy_at_the_same_timestamp`
- `rows_unique_to_the_older_side_survive_a_newer_partition_max`
- `streaming_diff_resolves_per_row_like_the_one_shot_form`

The last one exists because of the mistake above: it pins the production streaming path against the one-shot form. Reverting the streaming hunk alone fails it — verified by mutation, not assumed.

**1099 `ferrosa-cluster` tests pass. `fmt --check`, `clippy -D warnings`, `cargo doc -D warnings` clean.**

## Still to do

Not yet re-deployed to the cluster — the previous deploy is what exposed the streaming-path gap, so this needs the same end-to-end check before the release.


bkearns:
## Live verification on the native 3-node cluster

Deployed this branch merged with #339 and ran repair on the affected table.

**Before** (previous build, per-partition classification):

```
node1 = 78526 rows    node2 = 78529    node3 = 78529
node1 missing 3 rows, all in session 32ee0a9d
node1 empty shells: 29     node2: 25     node3: 25
repair: timestamp_ties 1, partitions streamed 0
```

(The 25 shells common to all three nodes are genuinely empty entities, not
divergence. node1's 4 extra shells are the contested rows.)

**After** (per-row reconciliation):

```
POST /api/cluster/repair?keyspace=agent_memory&table=entity_store
partitions_streamed_in: 1   partitions_streamed_out: 1   timestamp_ties: 0

node1 = 78529 rows    node2 = 78529
```

All 7 rows verified individually against node1 and node2 at CL=ONE with a full
scan pinned per host — a point read is served by any replica and cannot show
what one node holds:

| entity | node1 before | node1 after | node2 |
|---|---|---|---|
| 0c16ec2d, b2ef3a13, b34790ab | absent | present, populated | present, populated |
| dd11fed6, ddf9e5c6, a64ce319, ec00ea62 | empty shell | present, populated | present, populated |

`timestamp_ties` went 1 → 0: there were no genuine conflicts, exactly as the
row-level analysis predicted.

## One session failed, and it should have

```
sessions_ok: 1   sessions_failed: 1
errors: ["... apply write: invalid data: agent_memory.entity_store
          (column \"created_at\", index 3): TimestampType expects 8 raw bytes
          but value provided 3072"]
```

This is not a regression in this PR — it is a **pre-existing schema divergence**
the tie no-op had been hiding. `system_schema.columns` for
`agent_memory.entity_store`:

- node1: 19 columns
- node2: 19 columns
- **node3: 11 columns** — missing every column added by later ALTERs
  (`description`, `description_embedding`, `tags`, `properties`, `content_hash`,
  `updated_at`, `scope`, `ingested_by_session`)

Regular-column ordinals therefore differ, so a row streamed to node3 decodes
against the wrong columns — 3072 bytes is 768 f32, an embedding landing on a
timestamp column. Repair refusing the write is the correct outcome; node3 cannot
be repaired until its schema is restored. Tracked separately as `t_433207d8`,
along with `t_3eb803f1` (how node1 acquired cell-less row markers at all).

node1↔node2 converged fully in the same run.

bkearns:
## Semantics changed: equal-timestamp conflicts now converge (deterministic LWW)

Decision from Ben: repair should resolve these like Cassandra rather than leaving replicas divergent.

Previously a row was recorded in `timestamp_ties` and **left alone** when both replicas held the same cell at the same timestamp with different values. That avoided silently discarding a write, but the price was permanent divergence — repair reported success while the replicas stayed split forever. Repair declining to converge replicas is repair declining to do its job.

The storage engine already resolves exactly this case: `merge_rows` applies cell-level LWW keyed by `(column, path)`, and `cell_supersedes` breaks an equal-timestamp tie on `(has_value, value_bytes, local_deletion_time)`. That is a **total order**, so every replica computes the same winner regardless of which side it calls "local" — which is what makes convergence safe rather than oscillating. Repair now delegates to that same merge and sends the merged row to whichever side does not already hold it.

`timestamp_ties` is kept, and now means: at least one row here was settled by comparing values rather than by timestamp. A non-zero count still tells an operator that writes collided at identical timestamps — worth attention even though repair settled it.

### Superseded tests

Three tests encoded the old contract and were deliberately updated, each carrying a comment recording what changed and why:

- `diff_partition_sets_equal_timestamp_different_content_is_a_tie`
- `a_conflicting_row_at_the_same_timestamp_is_still_a_tie`
- `local_executor_records_timestamp_ties_without_streaming` → now asserts the two replicas **converge** on the deterministic winner instead of both keeping their divergent values

### New tests

- `an_equal_timestamp_conflict_converges_both_replicas_on_one_row` — the losing replica alone receives the row, and what it receives equals what the other already holds. Re-sending to the winner would be pure write amplification.
- `the_equal_timestamp_winner_is_the_same_from_either_side` — the property that makes this safe. Each node runs the diff with *itself* as one side; if the merge favoured the left argument, every node would pick its own copy and repair would never converge.

1101 `ferrosa-cluster` tests pass. fmt, `clippy -D warnings`, `cargo doc -D warnings` clean.

### Note on the live cluster

The verification posted earlier ran **before** this commit, on the reconciliation-only build. Its result stands (node1 healed, `timestamp_ties: 0`), but it did not exercise the LWW path — there were no genuine conflicts on that cluster to exercise it with. The convergence property is covered by the tests above rather than by that deploy.

bkearns:
## CI failure was real, and found two defects

`Test + Coverage` failed on `divergent_replicas_converge_and_repair_is_idempotent` — the repair fuzz harness. My local runs used `cargo test --lib`, which does not build integration tests, so CI caught what I did not. I've since run the whole surface locally.

Reproduced deterministically from the seed CI printed; it's now checked in as a proptest regression.

### 1. Partition-level state was never propagated

```
replica 0 key [76, 116, 2, 7] converged to ts 913688 but LWW winner is ts 1000001
```

Reconciling only rows left the partition tombstone and static row behind. A partition deleted at a newer timestamp on one replica has nothing row-shaped to stream, so the previous commit flagged the divergence and streamed **nothing** — the replica without the tombstone stayed behind the winner forever while repair reported success.

That is the same class of silent non-convergence this PR exists to fix, reintroduced one level up. The generator produces partition deletions (`marked_for_delete_at: 803291`) and static rows; `entity_store` on the live cluster has neither, which is why the deploy verification looked clean.

`partition_with_rows` already carries the sender's deletion and static row, so a partition-state divergence now forces an exchange in **both** directions even when no rows need to move. Both directions matter: sending only the newer side leaves the sender un-updated and the next pass streams again — and the harness checks idempotence too.

### 2. `InMemoryRepairStore` did not model the apply path it stands in for

`apply_one` **replaced** the whole stored partition whenever the incoming copy had a newer max timestamp. The real `apply_partitions` writes each row through `engine.write` and the engine merges cell-wise; nothing swaps partitions.

That became load-bearing as soon as repair started streaming only the rows a replica is missing: a trimmed partition either clobbered rows the receiver held, or was dropped entirely for having a lower max timestamp. Either way the replicas could not converge, and the second pass kept streaming. The double now merges rows, takes the newer deletion and merges the static row — what the engine actually does.

### Also

`newest_partition_timestamp` is gone with its last caller: the diff decides per row and the store merges, so nothing compares whole-partition timestamps any more.

The P0 OOM guard also rejected an earlier revision of this work for copying row data on a scan path. Rather than add an allow entry — the allow file says that rule stays enforced precisely so new row clones fail — the copies were removed: the streaming path moves its owned partitions, and `partition_with_rows` carries partition state without needing a whole-partition copy. `p0-oom-audit --enforce` reports 0 findings with no new allow entries.

### Verification

Rebased onto `main` after #339 merged, so this applies on top of the Merkle fix.

All 36 `ferrosa-cluster` test targets pass (including `repair_fuzz`), `ferrosa-storage` passes, and all four CI guards are clean locally: `fmt --check`, workspace `clippy -D warnings`, `cargo doc -D warnings`, `check-unbounded-reads.py`, `p0-oom-audit --enforce`.

PR #350, opened 2026-08-21
chore: pin the Rust toolchain to latest stable (1.98.0)

bkearns:
## What

Pins the Rust toolchain to **1.98.0**, the current stable (`88d9e12ae`, 2026-08-18).

## Why

The suite was running **four different compilers**:

| Repo | Was |
|---|---|
| ferrosa | 1.96.1 |
| ferrosa-memory | 1.96.0 |
| ferrosa-dbaas | 1.97.1 |
| ferrosa-mobile | *unpinned* — whatever `rustup default` resolved to (nightly, locally) while its CI ran 1.97.1 |

`ferrosa-memory`'s existing pin comment already argued the case: a contributor one release behind hits "fails locally, green in CI". That argument now extends across repos — `ferrosa-mobile` will link `ferrosa-memory-sync` for the live WebRTC control channel, and two workspaces cannot link under two compilers.

Companion PRs pin the other three repos to the same version.

## Verification

`cargo fmt --check` clean on 1.98.0.

Clippy was still running when these were pushed. **This is a version jump with `clippy -D warnings`, so new lints are plausible** — CI is the authority here, and any findings should be refactored rather than silenced with `#[allow]`.

### DIFF
--- a/ferrosa-cluster/src/ring/mod.rs
+++ b/ferrosa-cluster/src/ring/mod.rs
@@ -77,7 +77,7 @@ impl TokenRing {
         }
 
         // Wrap around to the beginning of the ring
-        for (_, &node_id) in self.ring.iter() {
+        for &node_id in self.ring.values() {
             if seen.insert(node_id) && is_normal(node_id) {
                 result.push(node_id);
                 if result.len() >= rf {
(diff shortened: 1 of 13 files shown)


bkearns:
## Verification on 1.98.0 — needs a fix before merge

`cargo fmt --check` passes. **Clippy fails** on one new 1.98 lint, `clippy::chunks_exact_to_as_chunks`:

```
ferrosa-index/src/vector/quantized/container.rs:124:42
ferrosa-index/src/lib.rs:486:10
```

Both are `chunks_exact(N)` with a constant `N`; 1.98 wants `as_chunks`. Mechanical, and should be refactored rather than `#[allow]`-ed.

Verified against a freshly reinstalled 1.98.0 toolchain — an earlier run on a locally corrupted toolchain produced bogus failures elsewhere, so these two are real.

Same lint hits ferrosa-memory#214 (one site). ferrosa-dbaas#60 fails a different one (`result_large_err`, 6 sites). ferrosa-mobile#2 is clean.

PR #357, opened 2026-08-22
fix(accord): PreAccept stalled forever when the fan-out exhausted

bkearns:
Stacked on #356 (`fix/accord-readvote-quorum`). Merge that first.

## The bug

An LWT on a fully healthy RF=3 cluster fails `Accord quorum unavailable`
with every node up and every replica voting.

Instrumenting the PreAccept fan-out gave the decisive measurement:

```
peers=2 votes=2 no_votes=0 rpc_failures=0   -> decision still Pending
```

Nothing failed. Both peers voted. The round still reached no decision, so
this was never a quorum-*availability* problem — the defect is in how the
phase concludes.

## Root cause

`handle_preaccept_ok` transitions on exactly two conditions:

| condition | decision |
|---|---|
| `total >= fast_quorum` **and every response agrees** | `FastPathCommit` |
| `total >= slow_quorum` **and some response disagrees** | `NeedAccept` |

For RF=3 that is `fast_quorum = 3`, `slow_quorum = 2`. Two **agreeing**
votes satisfy neither arm and fall through to `Pending`.

Returning `Pending` there is correct *while a third response may still
arrive* — deciding at the slow quorum would abandon the 1-RTT fast path on
every transaction. The bug is that **nothing ever concluded the phase**
once the fan-out was exhausted and no further response could arrive. The
coordinator sat at `Pending` and the caller reported quorum unavailable.

The leaseholder path was unaffected: it pushes an implicit vote at
construction and so reaches `total = 3`. That is why this reproduced only
on non-leaseholder coordinators and looked intermittent.

## Fix

`AccordCoordinator::finalize_preaccept()` runs after the fan-out when the
decision is still `Pending`. At that point a fast quorum is provably
unreachable, so:

- a slow quorum voted → commit through the Accept phase
- below a slow quorum → stay `Pending`; the caller still fails loud

## Second defect: a failed self-vote was discarded in silence

The non-leaseholder self-vote matched `SmResponse::PreAcceptOK` with no
`else` arm. A local replica that failed to vote cost the round a vote it
was counting on while looking identical to a remote timeout — the same
silent-drop shape as the FileSyncWriter EISDIR bug. It now logs the
response it actually received.

Also carries the fan-out vote / no-vote / RPC-failure counters that
produced the measurement above, so a future failed round reports *why* it
failed rather than only *that* it failed.

## Verification

Live 3-node RF=3 native cluster, 90 `INSERT ... IF NOT EXISTS` spread
across all three coordinators.

**Before** (unfixed binary):

```
[BEFORE] applied=0/90 failed=90
     30  Unable to complete the operation against any hosts  (127.0.0.1:19042)
     30  Unable to complete the operation against any hosts  (127.0.0.1:19043)
     30  Unable to complete the operation against any hosts  (127.0.0.1:19044)
```

with `peers=2 votes=2 no_votes=0 rpc_failures=0` logged for every one.

After: see the comment below — being measured on the redeployed cluster.

## Tests

4 new unit tests on the pure decision function; 307 accord lib tests green.
One test I first wrote asserted an unreachable state — a disagreeing vote
already decides at the slow quorum and never reaches `finalize` — and was
replaced with a test that pins that split.

Closes t_5b44db24.

### DIFF
--- a/ferrosa-cluster/src/accord/coordinator.rs
+++ b/ferrosa-cluster/src/accord/coordinator.rs
@@ -294,6 +294,40 @@ impl AccordCoordinator {
         CoordinatorDecision::Pending
     }
 
+    /// Conclude the PreAccept phase when no further response can arrive.
+    ///
+    /// `handle_preaccept_ok` deliberately stays `Pending` while a fast quorum
+    /// is still reachable: with RF=3 the fast path needs all three replicas,
+    /// so two agreeing votes must keep waiting rather than spend a second RTT
+    /// on the Accept phase. That is correct *while responses are outstanding*.
+    ///
+    /// Once the fan-out is exhausted the fast path is provably unreachable.
+    /// If a slow quorum answered, the round can and must commit through the
+    /// Accept phase. Without this the coordinator sat at `Pending` forever and
+    /// the caller reported "Accord quorum unavailable" on a fully healthy
+    /// cluster — every LWT on a non-leaseholder RF=3 coordinator that lost a
+    /// single vote failed, even though two replicas had agreed.
+    ///
+    /// Returns `Pending` when fewer than a slow quorum responded (the round is
+    /// genuinely unavailable and the caller must fail loud) or when a decision
+    /// was already reached, in which case this is a no-op.
+    pub fn finalize_preaccept(&mut self) -> CoordinatorDecision {
+        if self.phase != CoordinatorPhase::PreAccepting {
+            return CoordinatorDecision::Pending;
+        }
+
+        if self.preaccept_responses.len() < slow_quorum_size(self.rf) {
+            return CoordinatorDecision::Pending;
+        }
+
+        self.phase = CoordinatorPhase::Accepting;
+        self.rtt_count = 1;
+        CoordinatorDecision::NeedAccept {
+            t: self.merged_t,
+            deps: self.merged_deps.clone(),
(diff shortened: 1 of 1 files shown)


bkearns:
## Live verification complete

Redeployed all three nodes of the native RF=3 cluster onto this build and
re-ran the same 90 `INSERT ... IF NOT EXISTS` across all three coordinators.

| | applied | failed |
|---|---|---|
| before | **0 / 90** | 90 |
| after | **90 / 90** | 0 |

The stall signature is gone. Zero `PreAccept round reached no decision`
warnings on any node after the fix went live.

One warning did fire on node3 at 14:37:29, while node1 was still restarting:

```
peers=2 votes=1 no_votes=0 rpc_failures=1 slow_quorum=2
```

That is the **correct** behaviour, not a recurrence: one replica was
genuinely down, only one vote arrived, and one vote is below the slow
quorum of 2 — so the round stayed `Pending` and failed loud exactly as it
should. The bug signature was `votes=2` reaching no decision, and that has
not appeared once since the deploy. The `slow_quorum` field in that line
also confirms node3 was running this build.

Checked the Accept phase for the same class of defect: `handle_accept_ok`
has a single unconditional threshold (`>= slow_quorum` → commit), so it has
no equivalent gap. The hole was specific to PreAccept's two-condition
structure.

PR #387, opened 2026-09-08
ci: build the macOS release binary on the self-hosted Mac

bkearns:
macOS minutes bill at **10x Linux**.

This job builds an unsigned `aarch64-apple-darwin` binary and uploads a tarball — no `codesign`, no `notarytool`, no keychain — so none of the constraints that keep the Apple *signing* jobs on GitHub's pool apply to it.

One file changed.


bkearns:
Closing: this cannot work, and the reason is worth recording.

The runners are now registered at **org** level (they were scoped to \`ferrosa-mobile\` alone), so `[self-hosted, macOS, ARM64]` resolves across the org. But the Default runner group has **`allows_public_repositories = false`**, and `ferrosadb/ferrosa` is **public** — so this job would queue forever rather than run.

That setting should stay off. Self-hosted runners on a public repo let a fork PR execute arbitrary code on the machine, and this runner is a workstation, not a disposable VM. Trading that for macOS minutes is not a good trade.

`ferrosa-streamer` (private) keeps the change — that was the larger saving anyway, since its mac job runs on every push and this one only runs on release.

PR #393, opened 2026-09-11
Run every ferrosa workflow on GitHub-hosted runners

bkearns:
## Why no release has shipped a macOS artifact

Four release runs are queued right now with `Build macOS aarch64` never assigned a runner:

```
queued  v2026.09.11.0545
queued  v2026.09.11.0440
queued  v2026.09.11.0009
queued  v2026.09.10.2349
```

Their Linux jobs finish in minutes on hosted runners; only this one hangs. Nothing publishes, so `ferrosa-installer/manifest.toml` is still pinned to **v2026.09.09.0442** — two days and one engine fix behind.

## Cause

**ferrosa is a public repository, and GitHub does not offer self-hosted runners to a public repo's workflows at an arbitrary ref** — only on the default branch. This workflow triggers on `push: tags: v*`, so every release run sits at `refs/tags/...`.

Evidence, gathered today:

| case | ref | dispatches? |
|---|---|---|
| ferrosa (public) install-smoke, driver-tests | `main` | ✅ runs on macbook-m5max |
| ferrosa (public) release | `refs/tags/*` | ❌ never assigned, runners idle |
| ferrosa-memory-**private** release | `refs/tags/*` | ✅ same label set, no trouble |

The group's `allows_public_repositories=true` admits the repo; it does not lift the ref constraint.

## What was tried first, and did not work

- **Restarting the runner.** It re-created its session and logged `Listening for Jobs`; jobs still queued.
- **Moving runners between groups.** Put the known-good `macbook-m5max-2` into `ferrosa-trusted-builders`; no change. (This emptied `ferrosa-private-builders` — worth reverting: `PUT /orgs/ferrosadb/actions/runner-groups/4/runners/625`.)
- **Removing the group's workflow restriction.** The group had `restricted_to_workflows=true` with every entry pinned to `@refs/heads/main`, which looked like an exact match for the symptom. Setting it to `false` changed nothing, including for a tag pushed fresh afterwards to get a clean dispatch decision. Worth noting for anyone tempted by the same theory — and GitHub does not accept wildcards like `@refs/tags/*` in that field, so it could not have been fixed by adding an entry either.

## The change

One line: `runs-on: [self-hosted, macOS, ARM64]` → `runs-on: macos-14`.

macOS minutes bill at 10x, which is why the rest of the fleet is self-hosted — so the comment records exactly why this job is the exception, and how to undo it: make ferrosa private, or move the release to a main-ref trigger. It runs only on a tag push, not per-commit.

## After merge

Re-tagging (or re-running a queued release) should publish a full artifact set for head, which unblocks repinning the installer manifest and cutting a DMG that carries the partition-key index fix.


bkearns:
## Extended: all seven routings, not just the release macOS job

Per review, ferrosa now uses GitHub-hosted runners everywhere rather than keeping one exception:

```
driver-tests.yml   1 job   [self-hosted, Linux, X64]   -> ubuntu-latest
install-smoke.yml  6 jobs  [self-hosted, Linux, X64]   -> ubuntu-latest
                           [self-hosted, macOS, ARM64] -> macos-14
release.yml        1 job   [self-hosted, macOS, ARM64] -> macos-14
```

Worth stating the security half plainly, since it is a better reason than the release being stuck: a **public** repo running workloads from arbitrary refs on a machine somebody owns is the arrangement GitHub's ref constraint exists to prevent. Moving off self-hosted is the right end state here regardless of the dispatch symptom.

The capnproto guard from #391 stays. On a hosted runner `sudo -n` succeeds, so it costs nothing, and it keeps these jobs safe to point back at a self-hosted box if ferrosa ever goes private.

## A bug this exposed in my own test

`tests/ci/test_install_smoke_workflow.py` (added in #391) has `jobs_routed_to_self_hosted()`, which searched each job block for the string `self-hosted`. After the routings moved to hosted runners, it **kept passing** — on the strength of a *comment* that still mentioned self-hosted.

A test written to notice that nothing routes to self-hosted any more, satisfied by prose about self-hosted. Exactly the silent-pass shape the rest of this work has been about, in the guard itself.

It now reads `runs-on:` / `runner:` values only, and is verified in both directions:

```
comment mentioning self-hosted -> []        (must be empty)
actually routed to self-hosted -> ['build'] (must find it)
```

The precondition is inverted to match reality — these jobs are expected to be hosted while ferrosa is public — and the stale comments that fooled it are corrected rather than deleted, because the sudo hazard they describe is real and explains why the guard is still there.

PR #396, opened 2026-09-11
feat(index): stream index reads in bounded memory — ordered walk, partition-granular tenant index, memory-mapped sidecars

bkearns:
The streaming follow-up to #394. **Stacked on #394**: the base is its branch until it merges, then I'll rebase this onto `main`. Six commits, readable one at a time.

## 1. Global index reads stream in row order (`390c9147`)

Index reads used to hold their whole result in memory:
- **Each node:** index postings were kept in arrival order, so a node deduplicated its memtable and sidecar postings through a set of every row it had seen.
- **Coordinator:** it did the same across replicas (`select_all` feeding a `HashSet`).
- **CQL:** it collected every match into a `Vec` and paged by re-slicing it. Paging a tenant's 101,848 rows rebuilt the whole result on every page.

Now:
- **Postings are kept in row order**, `(partition key, clustering)`, in every source. A node k-way merges its sources, drops duplicates by comparing each row with the previous one, and resumes strictly after a cursor.
- **The coordinator** sends the cursor to every replica and merges their streams in the same order.
- **CQL** serves a plain projection one bounded page at a time, using the server default page size when the request is unpaged, and resumes from a `(pk, ck)` cursor.
- **Builtin aggregates** fold as rows stream.
- **An intersection** is a merge-join holding one head per index.

## 2. A partition-key index is read a partition at a time (`676b1238`, t_c5bccc65)

The index machinery is row-granular. #394's predecessor, d8ccb2f3, reused it unchanged for partition-key columns, so a tenant read did one point read per row: ~101k single-row reads for 1,285 partitions. Now:
- A partition-key index posts `(pk, [])` once per partition.
- The walk streams each partition's rows in bounded chunks through the retried `read_limited_rows[_from]`.
- A cursor reopens its own partition.
- Row postings in sidecars written before this change are skipped once their partition has been streamed.

## 3. Sidecars are memory-mapped (`a80f51a4`, t_7ac6b0e3)

Every SSTable's every sidecar used to be read whole onto the heap and held for the SSTable's life. Opening and walking one 64k-posting sidecar peaked at **31 MB** of heap. Now:
- **Format v2** adds an offset table and a footer with a body CRC. It's validated in one pass at open, binary-searched in place, and entries decode as borrowed slices. The same walk now peaks **under 64 KiB**, whatever the file size.
- **Mapped pages are file-backed page cache**, which the kernel reclaims under memory pressure (a cgroup reclaims them before an OOM kill).
- **Writers never truncate a published file.** They stream to a temp file, fsync, then rename.
- **v1 files are converted at open** through the spilling `ExternalSorter`.
- **Flush maps the sidecars it writes.** It used to keep a heap copy of each beside the file.

## 4. Indexes stay complete across compaction and backfill (`e59eb146`)

These are two silent short-answer bugs that exist on `main` today. Each lasted until the next restart:
- **Compaction** swapped its output in with **no** sidecars. 20 indexed rows read as **0** after a compaction. The output's sidecars are now built by k-way merging the inputs' sidecars. Postings are keys, not offsets, so they stay valid.
- **`CREATE INDEX` over flushed data** wrote the backfilled sidecar to disk but never installed it. It read **0 of 15** rows while the tracker called the index current. The scheduler now installs each sidecar before marking the SSTable indexed, and a failure marks the build failed.

## 5. S3 (`837e5120`)

- **Restore** (S3 bootstrap and snapshot/PITR restore) now pulls every index artifact of a generation from S3 completely before publishing it. Previously a node rebuilt from S3 had no indexes over restored data.
- **Late sidecars:** a sidecar built after its generation was already uploaded is now uploaded by the next S3 sync.
- **Gauges:** `ferrosa_storage_index_sidecar_mapped_{bytes,files}`.

## Tests (each written failing first)

- **Storage:** memtable and sidecar row order; the ordered merge and resume, plus a tripwire test against result-sized collections; partition postings and chunked reads (12 point reads → 0); mid-partition resume; legacy postings.
- **Memory bound:** a heap-bounded sidecar walk.
- **Sidecar lifecycle:** v1 conversion; rewriting while a reader has the file mapped; flush maps its sidecars; compaction (0 → 20); backfill (0 → 15).
- **S3:** both restore paths; the late upload; the gauges.
- **Cluster:** the ordered replica merge with RF>1 duplicates, error propagation, and a coordinator tripwire; a resumed scatter-gather across three real engines.
- **CQL:** bounded default pages and cursor walks; intersection paging; aggregates folding with zero rows collected; tenant pages cut mid-partition.
- **SPARQL:** object scans without an object index.

## Local verification

- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings` and `RUSTDOCFLAGS=-D warnings cargo doc --no-deps --workspace` are clean.
- `cargo test --workspace --no-fail-fast`: **7,325 passed**. Nine failed, and CI already excludes all of them:
  - `disk_fail_no_phantom_commits` is on CI's skip list (tracked as t_8e27ae39).
  - The other eight are in `ferrosa-jepsen` and `ferrosa-loadgen`, which CI excludes.

## Not in this PR (tracked)

**Other index reads that still collect results:**
- `ORDER BY`, `DISTINCT` and non-builtin functions over an index: t_d4bb3cc8
- full-text reads: t_91b6fbfc
- the keyed in-partition read: t_c6195429
- geo reads: t_0471348c

**Sidecar follow-ups:**
- remote-built sidecars are not pulled local: t_7b8ee7e9
- a not-yet-current index is still read, and can answer short: t_edd3be70
- orphaned sidecars are never deleted: t_71fc5111
- no cap on open mappings, and the late-upload queue is memory-only: t_77f65e8f
- index builds still hold every entry in a `Vec`: t_fb10fb7b
- regular-column indexes still do one point read per row: t_ffcacfbc

Refs t_50c8bc7d, t_c5bccc65, t_7ac6b0e3.


bkearns:
Superseded by #403, which carries every commit from this branch unchanged — including `05d7d2fe` and `ecb62662` from `agent/pr396-green` — and then fixes the flush path this PR left copying.

The read side here streams in O(1), but each flush still wrote its sidecars by copying every posting three times (`MemtableIndex::iter` deep-copied the tree, the flatten cloned the key per posting, `SidecarWriter::write` took its own `to_vec` to sort), plus a fourth copy on the image and error paths. #403 streams the memtable index straight into the sidecar instead, from a pinned snapshot, and proves it with a write-side budget in `sidecar_memory_bound.rs`.

#403 also folds in #401's rule and extends it to t_12457d3e: an index absent from this node's table no longer fails the read with a 500 — it is withheld and the licensed scan runs.

Nothing from this branch is lost; the branch itself is untouched. If you have work in flight on `agent/pr396-green`, rebase it onto `feat/streaming-index-v2`.

PR #401, opened 2026-09-12
fix(cql): stop reading a secondary index whose build has not finished

bkearns:
**This fixes a failure on `main`.** The Driver Smoke workflow is red on main today, and has been alternating between tests:

| run | branch | test | result |
|---|---|---|---|
| 00:41 | main | `test_hash_low_cardinality` | 2374 of 2500 rows |
| 00:27 | #396 | `test_btree_equality_lookup` | 0 of 1 rows |
| 23:15 | #396 | `test_hash_low_cardinality` | 2126 of 2500 rows |

`test_2i_demo.py` compares an indexed lookup against the `ALLOW FILTERING` scan of the same data. The indexed side keeps coming back short, and which test loses the race varies per run.

## The bug

`CREATE INDEX` over rows that already exist backfills asynchronously. Until it finishes the index covers part of the table, and the global index arm consulted it anyway, reporting that fraction as the whole answer — with nothing to distinguish it from a genuinely empty result.

The engine already knew better. Every build is `mark_pending`-ed before it is submitted, and the comment on that line says it is done so "a consult that finds the index empty can only trust it once the backfill has completed". `index_is_current` exists and the keyed-partition arm consults it. The global arm never did.

## The fix

An index whose build is unfinished is withheld from the planner, the same way an index that can never be built is withheld (d4aca15e), and for the same reason: slower and correct beats fast and wrong.

- With `ALLOW FILTERING`, the read takes the scan it already licensed.
- Without one, it is refused, naming the index and saying it is still building. Returning the partial answer is the silent short answer this line of work exists to remove, and a retry a moment later succeeds.

Two cases stay distinct. An index the tracker has never heard of is not mid-build — the engine refuses to read an index it does not have, naming it (t_50c8bc7d, #394). Leaving that to its own path keeps a lost index loud instead of quietly downgrading it to a scan; its regression test still passes unchanged. The tracker is also local, so the check applies only where the local engine is the whole picture, mirroring what the keyed consult already does.

## Relationship to #396

#396 fixes the same class of bug at a different layer, and the two are complementary rather than competing:

| | #396 | this PR |
|---|---|---|
| layer | storage — `read_by_index_each_after` errors when the index is not current | planner — the index is withheld from plan selection |
| with `ALLOW FILTERING` | errors | takes the scan the query already licensed |
| without one | errors | errors, naming the index |
| covers a replica serving an index read | yes | no — the tracker is local (t_f29318c4) |

Layered, the planner never reaches storage while a build is unfinished, a licensed scan keeps working, and #396 remains the backstop for callers that bypass the planner.

#396 also repairs `wait_for_index_build` in the driver test, which documented itself as polling until the index was functional while its body was `time.sleep(1)` — that is what let the race reach CI. This PR touched the same helper and has since dropped that edit (d933bc77) so the two cannot collide; either PR can land first.

**#396 is also the other half of the fix on main.** Driver Smoke on this branch still fails at 1621 of 2500 rows, with the tracker reporting the index *current* while it is short. That is the missing sidecar install (e59eb146, in #396 and not on main): the build "completes" over nothing. This PR removes the read-during-build race; #396 removes the missing install. Main needs both, which is why the same test passes on #396 and fails here.

## Tests (written failing first)

- `a_still_building_index_does_not_answer_a_read_short` — reproduces the CI symptom exactly, failing with "must not turn it into 0 rows" before the fix.
- `a_still_building_index_refuses_a_read_it_cannot_serve` — the unlicensed read is refused, naming the index.

Both use a new `setup_with_index_backend(IndexBackendConfig::Off)`, which builds no scheduler, so `CREATE INDEX` marks the flushed SSTable pending and nothing ever indexes it. That holds the real backfill window still rather than racing it.

## Local verification

- `cargo fmt --check`, `cargo clippy --all-targets -- -D warnings`, `RUSTDOCFLAGS=-D warnings cargo doc --no-deps --workspace` and `scripts/check-unbounded-reads.py` all clean.
- The `ci.yml` test command: **6,813 passed**. The only failures are five `ferrosa-udf` AssemblyScript tests that panic on an unset `FERROSA_ASC_BUNDLE`; CI builds that bundle in a prior step.

## Not in this PR (tracked)

The tracker is per-node, so a coordinator fanning out to replicas still cannot tell a replica's index is mid-build. Captured as t_f29318c4.

Refs t_edd3be70.


bkearns:
Note for anyone following this from #396: the rule this PR added has been **generalised on top of it** in #403, not reverted.

`index_build_has_finished` became `index_cannot_answer_completely`. It still withholds an index whose backfill is unfinished, and now also withholds one that is absent from this node's table though the schema lists it. That second case is t_12457d3e: this PR deliberately left it refusing to preserve #394's contract, and the ferrosa-memory CI log shows that refusal reaching clients as `Internal server error` on every entity stream, across `main` and every open PR of that repo:

```
secondary index 'idx_entity_by_tenant' is not declared on this node's table, so it
cannot answer; refusing to report its absence as zero matching rows
```

A licensed scan is the correct answer there, so #394's test now asserts the fallback and keeps the refusal for the query that licensed no scan. Both tests from this PR are in #403 unchanged.

PR #417, opened 2026-09-17
Fix role password persistence across restart

bkearns:
## Executive Summary

Changing the `cassandra` or `ferrosa_admin` password appeared successful, but standalone role DDL only changed the in-memory schema and cold start seeded defaults before reading durable auth state. A restart therefore restored the public default credentials. This change writes direct create/alter/drop operations to `system_auth.roles`, replays commit-log mutations before reconstructing roles, restores persisted roles before seeding missing defaults, and clears the built-in must-change marker after rotation.

Fresh installs no longer create an implicit `cassandra/cassandra` superuser. The legacy `cassandra` role is created only when an operator explicitly supplies `FERROSA_SUPERUSER_PASSWORD`; the normal seeded administrator remains `ferrosa_admin`. Existing persisted roles continue to replay normally.

Seeded plaintext roles now also carry a persisted SCRAM-SHA-256 verifier, so `ferrosa_admin` and `ferrosa_user` can authenticate through the Postgres listener instead of failing as `UnknownRole`.

## Changes

- Persist standalone `CREATE ROLE`, `ALTER ROLE`, and `DROP ROLE` mutations through `SystemTableWriter`.
- Store and reload bcrypt hashes and SCRAM verifiers from `system_auth.roles`.
- Replace fresh-process bootstrap roles with durable role records during cold start.
- Replay pending commit-log mutations before system-table reconstruction.
- Seed only roles absent after recovery and clear `must_change_password` after rotation.
- Require an explicit secret before creating the legacy `cassandra` superuser.
- Update protocol fixtures, example configuration, and crate documentation for the seeded-role contract.

## Validation

- Exact CI-equivalent all-feature workspace suite: 7,031 passed, 0 failed, 9 configured skips.
- `cargo clippy --workspace --all-targets --all-features -- -D warnings`: passed.
- Workspace rustdoc with warnings denied: passed.
- `ferrosa-schema`: 382 passed.
- `ferrosa-flight`: 46 passed.
- Focused graph range-read memory-bound suite: 5 passed.
- Cluster replay, CQL blocking-auth, and graph superuser authorization regressions: passed.
- Formatting, diff validation, and targeted pre-commit hooks: passed.


bkearns:

PR #418, opened 2026-09-17
Fix authenticated CQL v5 framing for Python clients

bkearns:
Authenticated native-protocol v5 connections switched to checksummed framing one response too late. After the unframed `AUTHENTICATE`, Python-driver and cqlsh enabled their segment decoder, read Ferrosa's raw `AUTH_SUCCESS` response byte `0x85` as a segment header, and failed with `CrcMismatchException`.

This moves the v5 transition to the response-to-STARTUP boundary required by the protocol. With authentication enabled, `AUTH_RESPONSE` and `AUTH_SUCCESS` now use modern checksummed frames. It also adds a raw-TCP boundary regression and a real Python-driver v5 authentication smoke test.

Validation:
- CQL all-features library and integration suite: 1,200 passed, 0 failed, 4 skipped
- Exact CI workspace test command: 7,029 passed, 0 failed, 9 skipped
- Workspace all-targets/all-features clippy
- OOM audit
- Release build for `ferrosa` and `ferrosa-ctl` with all features
- Formatting and targeted pre-commit checks
- Python test syntax compile
- Docker smoke configuration could not run locally because the Podman compatibility socket is stopped; CI runs the container case


bkearns:

PR #420, opened 2026-09-17
Prevent collection appends from panicking SSTable flush

bkearns:
A normal collection insert stores a legacy whole-value cell, while list/set/map updates now emit path-keyed element cells. When both reached the same memtable row, flush enabled complex SSTable encoding and then panicked on the live pathless cell; cell ordering could also trip the writer's sorted-column assertion. Shutdown consequently exited 101 and commit-log replay recreated the crash.

Memtable merge now normalizes a mixed row before reconciliation. It parses the existing whole-value wire blob without interpreting element types, expands it into a collection-deletion sentinel plus metadata-preserving element cells, and sorts the result by `(column, path)`. This runs for both memtable implementations and during commit-log replay. Malformed or allocation-amplifying blobs fail with `InvalidData` before mutating the existing row.

Regression coverage exercises:
- legacy list + append → flush → three-element readback
- legacy set + add → flush → two-element readback
- legacy map + put → flush → two-entry readback
- path-bearing cells on scalar columns fail loud

Validation:
- Exact CI workspace mirror with the CI AssemblyScript bundle: 7,030 passed, 0 failed, 9 infrastructure-gated skips
- `ferrosa-storage --all-features`: 1,165 passed, 0 failed, 6 ignored, excluding three tests that explicitly require external containers/captured SSTables
- `cargo clippy -p ferrosa-storage --all-targets --all-features -- -D warnings`
- Workspace rustdoc with `-D warnings`
- P0 OOM audit: 0 findings
- Formatting and targeted pre-commit hooks


bkearns:

PR #421, opened 2026-09-17
fix(cql): preserve prepared scalar wire types

bkearns:
The CQL EXECUTE path decoded date, time, duration, decimal, and varint bind values correctly, then converted them into parser literals that could not preserve their type or range. A prepared insert such as a bound date failed with `type mismatch: expected date, got integer literal`; large varints and decimal/duration values were rendered as debug strings.

This change keeps those values in their exact CQL wire encoding through AST substitution and decodes them against the prepared column type at the bridge boundary. It adds a Rust regression for all five types, including a varint larger than i64, plus a Python-driver smoke case that prepares, inserts, and reads every value back.

Validation:
- Ferrosa CQL suite: 1,186 passed, 0 failed, 4 skipped
- Exact CI all-features workspace gate: 7,029 passed, 0 failed, 9 skipped
- Workspace clippy with warnings denied
- Workspace rustdoc with warnings denied
- Targeted pre-commit hooks and Python syntax check


bkearns:

PR #422, opened 2026-09-17
Fix vector indexes created after existing rows

bkearns:
## Problem

Creating a vector index after rows were already written registered an empty live HNSW/HVQ index. The schema immediately made ANN queries select that index, so a query that returned correct brute-force results before `CREATE INDEX` returned zero rows afterward.

## Change

Build the new vector memtable index from the active and flushing memtables before publishing the updated store view. The backfill preserves each row's partition-key scope, which lets the indexed ANN path recover and rank the base partitions. Future writes continue through the existing incremental index maintenance path.

The CQL regression uses the reported order—insert rows, create the index, then query—and requires the exact nearest row. The storage regression covers both HNSW and HVQ.

## Validation

- Exact CI-equivalent all-feature workspace suite: 7,029 passed, 0 failed, 9 configured skips.
- `ferrosa-storage --lib`: 1,164 passed, 0 failed, 6 skipped.
- `ferrosa-cql --lib`: 1,121 passed, 0 failed, 3 skipped.
- Workspace clippy with all targets/features and warnings denied: passed.
- Workspace rustdoc with all features and warnings denied: passed.
- Formatting, diff validation, targeted pre-commit hooks, and secret scan: passed; the secret scan reported only pre-existing test/example credential patterns outside this diff.


bkearns:

PR #423, opened 2026-09-17
Fix count queries on system_schema.tables

bkearns:
`SELECT count(*) FROM system_schema.tables` emitted zero column metadata followed by ordinary table rows, so Python drivers decoded past the malformed frame and closed the connection. Route `system_schema.tables` through the aggregate-aware system-select encoder, retaining projection behavior and returning one bigint count row.

The full gate also exposed an ordering defect in the scheduler fairness check: its barrier synchronized thread starts but did not ensure all contenders had entered admission before the first scan consumed the shared budget. The check now waits until every scan is running or queued before measuring fairness.

Validation:
- CQL crate: 1,122 passed, 0 failed, 3 skipped
- focused scheduler fairness check: passed
- full CI-equivalent workspace suite: 7,029 passed, 0 failed, 9 skipped
- workspace clippy: clean
- workspace rustdoc: clean
- formatting: clean
- targeted pre-commit hooks: clean


bkearns:

PR #425, opened 2026-09-17
Support documented CQL transaction blocks

bkearns:
The documented single-query transaction form failed at the first body statement because the parser only accepted `BEGIN`, body statements, and `COMMIT` as separate requests. This change parses `BEGIN TRANSACTION; ...; COMMIT TRANSACTION;` and `ROLLBACK TRANSACTION;` into a bounded transaction block and executes every body statement through the existing registry-backed Accord path.

Body statements are limited to `SELECT`, `INSERT`, `UPDATE`, and `DELETE`; nested transactions, DDL, and explicit transaction IDs are rejected. Any body error rolls the transaction back immediately so the registry and connection shim do not leak state.

Validation:
- parser and router transaction-block regressions
- full `ferrosa-cql` suite
- exact all-features workspace CI test command: 7,032 passed, 0 failed, 9 skipped
- `cargo clippy --all-targets -- -D warnings`
- `cargo doc --no-deps --workspace` with warnings denied
- `cargo fmt --check`


bkearns:

PR #429, opened 2026-09-19
force_sync repairs eight bytes instead of rewriting the segment

bkearns:
`force_sync` called `force_full_flush`: `open(O_TRUNC)`, rewrite **every byte
accumulated in the segment**, `F_FULLFSYNC`. Inside one 32 MB segment that is
quadratic in the bytes already written — and `write_atomic_batch` calls it on
**every Accord apply**.

The entire rewrite existed to carry **one backwards write**. `write_sync_marker_at`
patches the previous marker so it points at the new one, and that patch lands
below the watermark an incremental flush writes forward from. The region is
`SYNC_MARKER_SIZE` — **eight bytes**.

So write the eight bytes with `pwrite` instead of the segment. Same durability,
same number of fsyncs: the repair rides inside the existing flush, before the
existing sync.

## The ordering that matters

The repair is decided **inside** the flush, under the file lock, using the
watermark that flush itself uses. Deciding it in the caller reads the watermark
before taking the lock, and a concurrent flusher can move it past the patched
marker in between — only the lock holder can tell whether those bytes already
reached disk.

`write_sync_marker_at` now returns the offset it actually patched, so a caller
cannot name a different marker if another writer moved the tracker between the
load and the store. The marker allocation is counted as an in-flight writer, the
way `append` counts an entry: a flusher snapshotting past the allocation must
wait for the bytes to exist, or it writes the hole and leaves an uncrossable
marker. The full rewrite masked that by resending everything.

## The test was itself tested

`force_sync_leaves_already_flushed_bytes_alone` plants a sentinel in the segment
**header** — a region the in-memory buffer disagrees with, and which no
incremental flush and no marker repair ever writes. A full rewrite restores the
header from the buffer and erases it.

Reverting `force_sync` to `force_full_flush` turns that test **red**
(`left: 0, right: 255`) and leaves the other two green. It isolates exactly this
behaviour and nothing else.

Deliberately **not** the `FULL_FLUSHES_TOTAL` counters: they are process-global
and every other test in the binary moves them concurrently, so any delta is
noise — the same trap that made an unrelated descriptor-counting test flaky.

- 1174 lib tests, all integration binaries: pass
- `fmt --check`, `clippy --all-targets -D warnings`: clean

## Found along the way, and NOT fixed here

**Concurrent `force_sync` callers corrupt the marker chain and lose entries.**
4 threads × 25 rounds replays **4 of 100**. Each caller picks its predecessor
independently, so links are lost or written backwards, and `SegmentReader` stops
at `next <= marker_offset`.

It is **pre-existing**: the full-rewrite baseline fails 13/25 runs, this version
8/25, identical signature. Reachable in production through `write_atomic_batch`.

Repro and analysis in `t_96008240` (p90). The failing test is deliberately
uncommitted so it cannot break the nightly `--ignored` job.

## Also

`force_full_flush` is left in place, unused, with a note. `t_2f26493e` decides
whether it and its now structurally-zero `FULL_FLUSH*` metrics go.

One reasoned-but-unverified piece: the in-flight-writer tracking on the marker
allocation closes a real hole but the window is too narrow to hit in a test —
removing it leaves the concurrency test passing 10/10. It costs a spin-wait of a
few nanoseconds. Happy to drop it if you'd rather not carry untested code.


bkearns:
## Do not merge yet — this PR opens a crash-window durability hole that `main` does not have

Found while fixing a *different* marker-chain defect (`t_96008240`) on top of this branch. Captured as **`t_0880ea84`**.

This is not the bug this PR fixes, and not a bug in the 8-byte repair itself. It is a consequence of narrowing the write: **each flush now repairs exactly one marker — the predecessor its own call patched** — where `main` rewrote the whole segment and therefore carried *every* pending patch out on every call.

### The window

`marker_patch_to_rewrite(prev, already_flushed)` returns `Some` only when `prev < already_flushed`. A predecessor above the watermark needs no repair, because the forward write is expected to cover it. With two concurrent `force_sync` callers, the flushes can complete in the opposite order to the links:

```
state: last marker P, already flushed (last_flushed >= P+8)

T1 links A:  buffer has A(EOF), P->A          (A > last_flushed)
T2 links B:  buffer has B(EOF), A->B

T2 flushes FIRST: forward-writes buf[last_flushed..pos] — covers A and B.
  marker_patch_to_rewrite(A, last_flushed) is None, because A >= last_flushed.
  Disk: P still UNPATCHED, A->B, B->EOF.  fsync.  force_sync returns Ok.

  *** crash here ***
  P->A exists only in the buffer. SegmentReader walking from P cannot reach
  A or B. Every entry T2 was just told was durable is unreplayable.

T1 then flushes: P < last_flushed -> repairs P->A. Chain complete.
```

**Severity:** acknowledged-durable data lost on crash, on the Accord apply path (`engine.rs` ~6906).
**Probability:** needs two concurrent `force_sync`s *and* a crash inside a sub-millisecond window.

### Why it is not visible in the tests

`concurrent_force_sync_keeps_every_entry_replayable` asserts *after* all `force_sync` calls have returned, when the final state is correct — so it passes. There is no fault-injection harness in `ferrosa-storage` today, and catching this needs one that stops the process between T2's fsync and T1's repair.

### The fix, and why it is not inline

Each flush should carry every marker patch not yet on disk, not only its own. Markers are monotone so the pending set is a contiguous run, and a low-watermark plus a walk of the in-buffer chain would find it — but that walk reads 8-byte links another thread may be mid-writing under `marker_link` while the walker holds only the file lock. Torn reads.

Doing it correctly means recording the pending set explicitly (offsets under `marker_link`, drained under the file lock). That is materially more machinery than either this PR or the `t_96008240` fix, and belongs in its own review.

The performance argument for this PR is unchanged and still right — a 32 MB rewrite to patch 8 bytes, per Accord apply, is what drives the measured ~15,000 IOPS at ~5 KB for ~60 MB/s. This needs to land. It should land **with** the pending-patch set, or knowingly with the window documented and `t_0880ea84` scheduled.

